Issue #37, opened 2020-03-26
[proposal] Versioning and deprecation


gundalow:
# SUMMARY

* In ansible/ansible:

  * There is a single version number which is over everything shipped in Ansible
  * Doesn't use semver, uses X.Y (ie 2.9) as the major number
  * Deprecations are done over 4 versions (~ 2 years)
* In Collections:

  * Can be versioned and released independently to Ansible
  * MUST use `semver (Semantic Versioning) <https://semver.org/>`_

# Use-cases

## Users

- As a user I want to know if an upgrade to a Collection may including breaking changes
- As a user I want to be able to install a newer Collection, even if it's included in ACD
- As a user I want to know if a collection will work with the version of ansible I’ve standardized on

## Contributors

- As a Contributor I may want to take advantage of new functionality in ansible-base, and therefore set dependencies on that
- As a Contributor I want to know when I can introduce breaking changes, and when I can't (freeze window)
- As a Contributor I want to specify the required version of Ansible

## Maintainers

- As a Maintainer I want to release a Collection without an Ansible version increment
- As a Maintainer I want to release a Collection without an ACD version increment
- As a Maintainer I want to detail which minimum versions of Ansible my Collection requires

## Community Team

- As a member of the Community Team, we need to know which commit of a Collection to pull into ACD

## Deprecation Cycle

- As a user I want to know that Deprecation notices WILL be honored
- As a maintainer I want to know when I can act on the removed_in_version: 2.14 notices previously given 

# Proposal Versioning

- All collection owners should follow semver versioning standards for versioning their collections.
  - https://semver.org/
    - Given a version number `MAJOR.MINOR.PATCH`, increment the:
      - `MAJOR` version when you make incompatible API changes,
      - `MINOR` version when you add functionality in a backwards compatible manner, and
      - `PATCH` version when you make backwards compatible bug fixes.
    - Versioning should be based on APi changes of collections not API changes of ansible-base
    - Example:   The first collection created for some content can start with the version number like  `1.0.0`
- Galaxy and AH tooling will only enforce semver at the import time to make sure proper versioning standards are defined and followed.  There will be no validation if versioning is followed in principles of API changes in the collection.
- Ansible Collection owner has to specify metatag in collection that will specify which ansible-base version and above the collection is supported e.g. meta/($somefile_tbd).  It will have some line like  `requires_ansible: '>=2.10'` - implemented by [Collection redirection/tombstoning #67684](https://github.com/ansible/ansible/pull/67684)
[...]

PR #56, opened 2020-05-03
Add code for Ansible Galaxy release automation for Collections

ericsysmin:
##### SUMMARY
Add documentation for Ansible Galaxy release automation via GitHub Action Workflow

##### ISSUE TYPE
- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/galaxy_releases.rst
+++ b/docs/galaxy_releases.rst
@@ -169,7 +169,7 @@ new file :code:`release.yml` and it should look something like this.
   jobs:
     release:
       name: Build and release a collection
-      runs-on: ubuntu-latest
+      runs-on: ubuntu-18.04
       env:
         ANSIBLE_FORCE_COLOR: 1
       steps:
--- a/docs/galaxy_releases.rst
+++ b/docs/galaxy_releases.rst
@@ -1,0 +1,218 @@
+******************************
+Ansible Galaxy Release Process
+******************************
+
+.. contents:: Topics
+
+Overview
+========
+In this document we will cover common recommended ways to automate your Ansible
+Collection release process.
+
+**Warning:** Subject to frequent updates
+       This is a "living document", expect it to change as we progress with the
+       Collections work over the next few months.
+
+GitHub Action Workflow for Collection Release
+=============================================
+
+Ansible Galaxy does not allow us to use the web-hook we have grown accustom to
+for publishing or releasing an Ansible Collection. To publish your Ansible
+Collection to GitHub you have two options:
+
+#.  Uploading the :code:`tar.gz` that was created during the build command
+    :code:`ansible-galaxy collection build`
+#.  By using the :code:`ansible-galaxy collection publish` command after the
+    build command.
+
+Ansible Collection Build Process
+--------------------------------
+
+
+Create the :code:`build/` directory.
+::
+
+  build
+  ├── galaxy_deploy.yml
+  └── templates
+      └── galaxy.yml.j2
+
+Instead of having the :code:`galaxy.yml` file in our root, we will be generating
+this each time we create a release.
+
+You will need to create the :code:`galaxy_deploy.yml` Ansible playbook which
+will handle the build and publish steps for us. It's contents should look
+something like this.
+
+galaxy_deploy.yml
+^^^^^^^^^^^^^^^^^
+.. code-block:: yaml
+
+  ---
+  - hosts: localhost
+    connection: local
+    gather_facts: false
+    vars:
+      # This should be set via the command line at runtime.
+      tag: "{{ github_tag.split('/')[-1] }}"
+    pre_tasks:
+      - name: Ensure the ANSIBLE_GALAXY_TOKEN environment variable is set.
+        fail:
+          msg: ANSIBLE_GALAXY_TOKEN is not set.
+        when: "lookup('env','ANSIBLE_GALAXY_TOKEN') | length == 0"
+      - name: Ensure the ~/.ansible directory exists.
+        file:
+          path: ~/.ansible
+          state: directory
+      - name: Write the Galaxy token to ~/.ansible/galaxy_token
+        copy:
+          content: |
+            token: {{ lookup('env','ANSIBLE_GALAXY_TOKEN') }}
+          dest: ~/.ansible/galaxy_token
+    tasks:
+      - name: Template out the galaxy.yml file.
+        template:
+          src: templates/galaxy.yml.j2
+          dest: ../galaxy.yml
+        register: galaxy_yml
+      - name: Build the collection. # noqa 503
+        command: >
+          ansible-galaxy collection build
+          chdir=../
+        when: galaxy_yml.changed
+      - name: Publish the collection. # noqa 503
+        command: >
+          ansible-galaxy collection publish ./namespace-collection-{{ tag }}.tar.gz
+          chdir=../
+        when: galaxy_yml.changed
+
+You'll then need to create the :code:`build/templates` directory which we will
+be placing our :code:`galaxy.yml.j2` jinja template. Please edit the jinja
+template to your liking. Do not change the :code:`version` section as this is
+necessary for Ansible to properly template your :code:`galaxy.yml` for the
+specific release.
+
+galaxy.yml.j2
+^^^^^^^^^^^^^
+.. code-block:: yaml
+
+  ### REQUIRED
+
+  # The namespace of the collection. This can be a company/brand/organization or product namespace under which all
+  # content lives. May only contain alphanumeric lowercase characters and underscores. Namespaces cannot start with
+  # underscores or numbers and cannot contain consecutive underscores
+  namespace: namespace_name
+
+  # The name of the collection. Has the same character restrictions as 'namespace'
+  name: collection_name
+
+  # The version of the collection. Must be compatible with semantic versioning
+  version: "{{ tag }}"
+
+  # The path to the Markdown (.md) readme file. This path is relative to the root of the collection
+  readme: README.md
+
+  # A list of the collection's content authors. Can be just the name or in the format 'Full Name <email> (url)
+  # @nicks:irc/im.site#channel'
+  authors:
+    - "Author1"
+    - "Author2 (https://author2.example.com)"
+    - "Author3 <author3@example.com>"
+
+
+  ### OPTIONAL but strongly recommended
+
+  # A short summary description of the collection
+  description: Description of what this collection has
+
+  # Either a single license or a list of licenses for content inside of a collection. Ansible Galaxy currently only
+  # accepts L(SPDX,https://spdx.org/licenses/) licenses. This key is mutually exclusive with 'license_file'
+   license:
+     - GPL-2.0-or-later
+
+  # The path to the license file for the collection. This path is relative to the root of the collection. This key is
+  # mutually exclusive with 'license'
+  # license_file: ''
+
+  # A list of tags you want to associate with the collection for indexing/searching. A tag name has the same character
+  # requirements as 'namespace' and 'name'
+  tags: []
+
+  # Collections that this collection requires to be installed for it to be usable. The key of the dict is the
+  # collection label 'namespace.name'. The value is a version range
+  # L(specifiers,https://python-semanticversion.readthedocs.io/en/latest/#requirement-specification). Multiple version
+  # range specifiers can be set and are separated by ','
+  dependencies: {}
+
+  # The URL of the originating SCM repository
+  repository: https://github.com/my_org/my_collection
+
+  # The URL to any online docs
+  documentation: https://github.com/my_org/my_collection
+
+  # The URL to the homepage of the collection/project
+  homepage: https://github.com/my_org/my_collection
+
+  # The URL to the collection issue tracker
+  issues: https://github.com/my_org/my_collection/issues
+
+
+Configure GitHub Workflow
+-------------------------
+
+To be able to execute the Ansible playbook we need to configure the GitHub
+Actions Workflow. To do this we will need to create the
+:code:`.github/workflows/` folder if it does not already exist. Then create a
+new file :code:`release.yml` and it should look something like this.
+
+.. code-block:: yaml
+
+  name: "release"
+  on:
+  release:
+    types:
+      - created
+  jobs:
+  release:
+    runs-on: ubuntu-18.04
+    env:
+      ANSIBLE_GALAXY_TOKEN: ${{ secrets.ANSIBLE_GALAXY_TOKEN }}
+      ANSIBLE_FORCE_COLOR: 1
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ansible
+
+      - name: Run role test
+        run: >-
+          ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}"
+
+In this file you may notice we have :code:`github.ref`, and
+:code:`secrets.ANSIBLE_GALAXY_TOKEN`vars. These vars will be handled by GitHub
+on execution of the workflow. The :code:`github.ref` is automatic and will be
+parsed from GitHub as it passes this value on release to the Workflow
+automatically.
+
+Configure ANSIBLE_GALAXY_TOKEN Secret
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+#.  Go to Ansible Galaxy and get your API Key. You can find it on
+    https://galaxy.ansible.com/me/preferences in the section API Key. Just click
+    on "Show API Key".
+#.  Go to your GitHub repository. Click on "Settings" -> "Secrets".
+#.  Click on "Add a new secret"
+#.  For the name of the secret, it's important you use ANSIBLE_GALAXY_TOKEN
+    as that is what the Workflow is referencing. These values must match
+#.  After you place the API Key as the value, click on "Add secret".
+
+After this step is completed you should be able to trigger a release and watch
+your Ansible collection be published on the GitHub Actions page.
--- a/docs/galaxy_releases.rst
+++ b/docs/galaxy_releases.rst
@@ -55,36 +55,28 @@ galaxy_deploy.yml
     vars:
       # This should be set via the command line at runtime.
       tag: "{{ github_tag.split('/')[-1] }}"
-    pre_tasks:
-      - name: Ensure the ANSIBLE_GALAXY_TOKEN environment variable is set.
-        fail:
-          msg: ANSIBLE_GALAXY_TOKEN is not set.
-        when: "lookup('env','ANSIBLE_GALAXY_TOKEN') | length == 0"
-      - name: Ensure the ~/.ansible directory exists.
-        file:
-          path: ~/.ansible
-          state: directory
-      - name: Write the Galaxy token to ~/.ansible/galaxy_token
-        copy:
-          content: |
-            token: {{ lookup('env','ANSIBLE_GALAXY_TOKEN') }}
-          dest: ~/.ansible/galaxy_token
-          mode: '0600'
     tasks:
+      - name: Ensure that the ansible_galaxy_apikey exists
+        assert:
+          that:
+            - ansible_galaxy_apikey | length > 0
       - name: Template out the galaxy.yml file.
         template:
           src: templates/galaxy.yml.j2
           dest: ../galaxy.yml
         register: galaxy_yml
       - name: Build the collection. # noqa 503
-        command: >
-          ansible-galaxy collection build
-          chdir=../
+        command: ansible-galaxy collection build
+        args:
+          chdir: ../
         when: galaxy_yml.changed
       - name: Publish the collection. # noqa 503
         command: >
-          ansible-galaxy collection publish ./namespace-collection-{{ tag }}.tar.gz
-          chdir=../
+          ansible-galaxy collection publish
+            ./ericsysmin-test_collection-{{ tag }}.tar.gz
+            --api-key={{ ansible_galaxy_apikey }}
+        args:
+          chdir: ../
         when: galaxy_yml.changed
 
 You'll then need to create the :code:`build/templates` directory which we will
@@ -168,34 +160,34 @@ new file :code:`release.yml` and it should look something like this.
 
 .. code-block:: yaml
 
+  ---
   name: "release"
   on:
-  release:
-    types:
-      - created
+    release:
+      types:
+        - created
   jobs:
-  release:
-    runs-on: ubuntu-18.04
-    env:
-      ANSIBLE_GALAXY_TOKEN: ${{ secrets.ANSIBLE_GALAXY_TOKEN }}
-      ANSIBLE_FORCE_COLOR: 1
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v1
+    release:
+      name: Build and release a collection
+      runs-on: ubuntu-latest
+      env:
+        ANSIBLE_FORCE_COLOR: 1
+      steps:
+        - name: Check out code
+          uses: actions/checkout@v1
 
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v1
-        with:
-          python-version: 3.8
+        - name: Set up Python 3.8
+          uses: actions/setup-python@v1
+          with:
+            python-version: 3.8
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install ansible
+        - name: Install dependencies
+          run: pip install ansible
 
-      - name: Run the Ansible Galaxy deploy playbook
-        run: >-
-          ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}"
+        - name: Run the Ansible Galaxy deploy playbook
+          run: ansible-playbook -i 'localhost,' build/galaxy_deploy.yml
+            -e "github_tag=${{ github.ref }}"
+            -e "ansible_galaxy_apikey=${{ secrets.ANSIBLE_GALAXY_APIKEY }}"
 
 In this file you may notice we have :code:`github.ref`, and
 :code:`secrets.ANSIBLE_GALAXY_TOKEN` vars. These vars will be handled by GitHub
@@ -203,8 +195,8 @@ on execution of the workflow. The :code:`github.ref` is automatic and will be
 parsed from GitHub as it passes this value on release to the Workflow
 automatically.
 
-Configure ANSIBLE_GALAXY_TOKEN Secret
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Configure ``ANSIBLE_GALAXY_TOKEN`` Secret
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 #.  Go to Ansible Galaxy and get your API Key. You can find it on
     https://galaxy.ansible.com/me/preferences in the section API Key. Just click
--- a/docs/galaxy_releases.rst
+++ b/docs/galaxy_releases.rst
@@ -69,6 +69,7 @@ galaxy_deploy.yml
           content: |
             token: {{ lookup('env','ANSIBLE_GALAXY_TOKEN') }}
           dest: ~/.ansible/galaxy_token
+          mode: '0600'
     tasks:
       - name: Template out the galaxy.yml file.
         template:
--- a/docs/galaxy_releases.rst
+++ b/docs/galaxy_releases.rst
@@ -197,7 +197,7 @@ new file :code:`release.yml` and it should look something like this.
           ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}"
 
 In this file you may notice we have :code:`github.ref`, and
-:code:`secrets.ANSIBLE_GALAXY_TOKEN`vars. These vars will be handled by GitHub
+:code:`secrets.ANSIBLE_GALAXY_TOKEN` vars. These vars will be handled by GitHub
 on execution of the workflow. The :code:`github.ref` is automatic and will be
 parsed from GitHub as it passes this value on release to the Workflow
 automatically.
--- a/docs/galaxy_releases.rst
+++ b/docs/galaxy_releases.rst
@@ -193,7 +193,7 @@ new file :code:`release.yml` and it should look something like this.
           python -m pip install --upgrade pip
           pip install ansible
 
-      - name: Run role test
+      - name: Run the Ansible Galaxy deploy playbook
         run: >-
           ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}"
 


docs/galaxy_releases.rst
```diff
@@ -0,0 +1,218 @@
...
+        when: "lookup('env','ANSIBLE_GALAXY_TOKEN') | length == 0"
+      - name: Ensure the ~/.ansible directory exists.
+        file:
+          path: ~/.ansible
+          state: directory
+      - name: Write the Galaxy token to ~/.ansible/galaxy_token
+        copy:
+          content: |
+            token: {{ lookup('env','ANSIBLE_GALAXY_TOKEN') }}
+          dest: ~/.ansible/galaxy_token
```
abadger, line 71: If you write this to disk, make sure the file's nose is 0600

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,218 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ansible
+
+      - name: Run role test
+        run: >-
+          ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}"
+
```
abadger, line 198: Is there a reason to have both a playbook and an action? Why not put everything into an action or create different instructions for people who want to do things differently

abadger:
One overarching question, is this still gplv3 compliant?  Is everything used to build the collection inside the collection tarball (i mean, the .Github directory and the deploy playbook)?

docs/galaxy_releases.rst
ericsysmin: The playbook does the templating and processing of the values. What action do you see here?

ericsysmin:
@abadger were not uploading the tar.gz created by GitHub, were uploading the asset made by Ansible Collection build command. But there is a tar.gz you could download just like you currently do on roles, or any release on GitHub. I just made it easier to tag the release the way you want. However currently Ansible build does include molecule, build, and other directories, not in excludes (supported in 2.10).

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,218 @@
...
+Create the :code:`build/` directory.
+::
+
+  build
+  ├── galaxy_deploy.yml
+  └── templates
+      └── galaxy.yml.j2
+
+Instead of having the :code:`galaxy.yml` file in our root, we will be generating
```
felixfontein, line 40: One disadvantage of this approach is that tests which need to know the next planned version number cannot read it from anywhere. Such tests don't yet exist in ansible-test, but I'm working on some (ansible/ansible#68646, ansible/ansible#66920). This would then mean that the version number must be supplied some other way in CI, and that it needs to be stored somewhere else.

ericsysmin: Couldn't we generate it then run ansible-test?

docs/galaxy_releases.rst
ericsysmin: Good point will make the change

docs/galaxy_releases.rst
felixfontein: How should ansible-test know the next released version of the collection if it can't read it from galaxy.yml?

ericsysmin: @geerlingguy have you thought about this? I know you're also using this process on one of your collections at the moment but with Travis-CI.

geerlingguy: Yeah... this is one of the annoying things that I don't know how we'd be able to solve it in a sane way. Basically you either have to tie your CI system back to git, so it can modify `galaxy.yml` then push back a commit—but that would happen _after_ you create the git tag.

Otherwise, for every release, the process becomes much more annoying (compared to roles, where you just run `git tag x.y.z && git push --tags` and that's it):

  1. Choose what you want your tag to be. Update `galaxy.yml` with that tag.
  2. Push that commit with the new tag in it... make sure CI passes.
  3. `git tag x.y.z && git push --tags`
  4. Wait for CI process to then build and publish the release.

Note that I've also opened [[Proposal] Support building and publishing collections server side](https://github.com/ansible/galaxy_ng/issues/133), which might also alleviate a lot of the annoyance over managing collections. Roles are/[[were](https://github.com/ansible/galaxy_ng/issues/58)?] much simpler to maintain.

felixfontein: Another (related) question: where in that process do you update the changelog / release notes? That should also contain the new version number (and more release-dependent information).

ericsysmin: When you create a release through GitHub. However, do we even have an official process for updating changelogs on collections yet?

felixfontein: I don't think we have. It's a related problem though, since the changelog also needs to know the release version, and after the release needs to be updated in some way so that new entries can be added for the next version.

I'm currently working on generalizing Ansible's changelog generator to collections (https://github.com/felixfontein/ansible-changelog/); it also uses galaxy.yml to determine the relesae version by default (you can still specify it though).

At least with this changelog generator, the release + changelog process could look like this:

0. (Not really part of the process: galaxy.yml contains the version of the next release; changes come with changelog fragments.)
1. Run `ansible-changelog release`, which creates a new release in the changelog with the current set of fragments. (Also adds a list of new modules/plugins when `version_added` is correctly used.)
2. Commit updated `changelog.yaml` file and generated .rst version of changelog. (The `.yaml` file is both metadata and what is used by ACD to compile combined changelog.)
3. Create tag.
4. Build and upload collection.
5. Update `galaxy.yml` for the next version and commit.

(And somewhen also push.)

This could be automated (by a script, playbook, github action, a combination of these, ...), except of course determining the next release version, which has to be done manually in some way.

ericsysmin: 1. Run Ansible playbook to generate galaxy.yml
2. Run ansible-test
3. Run ansible-changelog
4. Run ansible collection build
5. Run ansible collection publish

However,  by the time you even do a release isn't it expected you already had a branch with the release you're releasing from that has been being tested? Just a thought.

felixfontein: You might also want to run ansible-test *before* starting the release process :)

Also, you're missing a commit of the updated changelog in your steps.

> However, by the time you even do a release isn't it expected you already had a branch with the release you're releasing from that has been being tested? Just a thought.

That depends a lot on how the versioning/release process for the collection looks like. There will be different approaches, I guess, and right now that is also [under discussion](https://github.com/ansible-collections/overview/issues/37) - well, not that there is much of a discussion right now.

ericsysmin: I mean who starts a release without ansible-test, molecule, etc running before you create a release. Each PR and Push should be getting tested before any of these steps. Because I think it's relative to this discussion, please checkout these documents and pages.

https://github.com/geerlingguy/ansible-collection-php_roles/blob/master/.travis.yml
https://github.com/geerlingguy/ansible-collection-k8s/blob/master/.travis.yml
https://www.jeffgeerling.com/blog/2020/automatically-building-and-publishing-ansible-galaxy-collections
https://www.jeffgeerling.com/blog/2019/how-add-integration-tests-ansible-collection-ansible-test
https://ericsysmin.com/2020/05/02/ansible-collections-automating-the-release-process-to-galaxy/
https://ericsysmin.com/2020/04/30/ansible-collections-role-tests-with-molecule/

gundalow: @pabelanger how are you updating the version in galaxy.yml in Zuul?

docs/galaxy_releases.rst
abadger: I mean, the GitHub action. Although templating might be less easy with a raw action.

abadger:
> @abadger were not uploading the tar.gz created by GitHub, were uploading the asset made by Ansible Collection build command. But there is a tar.gz you could download just like you currently do on roles, or any release on GitHub. I just made it easier to tag the release the way you want. However currently Ansible build does include molecule, build, and other directories, not in excludes (supported in 2.10).

<Nod>. Yeah, we have to be careful not to exclude things used to build the asset.  Otherwise it no longer counts as source for the gpl.

docs/galaxy_releases.rst
ericsysmin: Let's see where the discussion ends up, maybe that ends up being where we take this now that people want all these other steps in place.

ericsysmin:
> > @abadger were not uploading the tar.gz created by GitHub, were uploading the asset made by Ansible Collection build command. But there is a tar.gz you could download just like you currently do on roles, or any release on GitHub. I just made it easier to tag the release the way you want. However currently Ansible build does include molecule, build, and other directories, not in excludes (supported in 2.10).
> 
> . Yeah, we have to be careful not to exclude things used to build the asset. Otherwise it no longer counts as source for the gpl.

Ah, then that's a problem for Ansible Collections in general as they do exclude directories by default.

docs/galaxy_releases.rst
felixfontein: @ericsysmin I mean running sanity tests which actually need to know the version number, like checking for deprecated features of modules, or making sure that PRs contain the correct value of `version_added`. That is of course not relevant for collections with roles in them (from the links I see that you seem to be more interested in these); for collections with modules, especially large and diverse ones like community.general and community.network, these tests are more important (IMO very important - for the same reasons that these tests existed and still exist for PRs in ansible/ansible itself).

pabelanger: @gundalow https://github.com/ansible-network/releases/blob/master/ansible_releases/cmd/generate_ansible_collection.py

we use pbr, to generate semver based on git commit. Then repopulate galaxy.yml with the version number

abadger:
Yep :-(. I want to make sure we do it right but I'm not sure what to do about other people yet.  (It's a problem that attends from Galaxy only hosting the built collections, not a source tarball)

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,218 @@
...
+  name: "release"
+  on:
+  release:
+    types:
+      - created
+  jobs:
+  release:
+    runs-on: ubuntu-18.04
+    env:
+      ANSIBLE_GALAXY_TOKEN: ${{ secrets.ANSIBLE_GALAXY_TOKEN }}
```
webknjaz, line 180: I'd rather open this token only to the upload step for security reasons

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,218 @@
...
+.. code-block:: yaml
+
+  name: "release"
+  on:
+  release:
+    types:
+      - created
+  jobs:
+  release:
+    runs-on: ubuntu-18.04
```
webknjaz, line 178: Probably better to use
```suggestion
    runs-on: ubuntu-latest
```

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,218 @@
...
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ansible
```
webknjaz, line 194: It's inconsistent with the previous command. And probably it's better to install under the user dir:
```suggestion
          python -m pip install ansible --user
          echo "::add-path::${{ env.HOME }}/.local/bin"
```

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+    pre_tasks:
+      - name: Ensure the ANSIBLE_GALAXY_TOKEN environment variable is set.
+        fail:
+          msg: ANSIBLE_GALAXY_TOKEN is not set.
+        when: "lookup('env','ANSIBLE_GALAXY_TOKEN') | length == 0"
+      - name: Ensure the ~/.ansible directory exists.
+        file:
+          path: ~/.ansible
+          state: directory
+      - name: Write the Galaxy token to ~/.ansible/galaxy_token
```
webknjaz, line 67: Why is it needed to write this on disk at all?

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+    tasks:
+      - name: Template out the galaxy.yml file.
+        template:
+          src: templates/galaxy.yml.j2
+          dest: ../galaxy.yml
+        register: galaxy_yml
+      - name: Build the collection. # noqa 503
+        command: >
+          ansible-galaxy collection build
+          chdir=../
```
webknjaz, line 82: this chdir thing looks weird. did you mean something like
```suggestion
        args:
          chdir: ../
```

docs/galaxy_releases.rst
webknjaz: Even better if you could pass it via the args of the ansible-playbook CLI command (all the secrets are automatically sanitized in GitHub Workflows output)

docs/galaxy_releases.rst
webknjaz: Yeah, it could be a great idea to actually create our own action. But I'd not mix build&publish, because one should inject some testing in between, so it'd be two actions.

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ansible
+
+      - name: Run the Ansible Galaxy deploy playbook
+        run: >-
+          ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}"
```
webknjaz, line 198: I think I'd rather pass the secret like this:
```suggestion
          ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=${{ github.ref }}" -e "ansible_galaxy_token=${{ secrets.ANSIBLE_GALAXY_TOKEN }}"
```

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+
+In this file you may notice we have :code:`github.ref`, and
+:code:`secrets.ANSIBLE_GALAXY_TOKEN` vars. These vars will be handled by GitHub
+on execution of the workflow. The :code:`github.ref` is automatic and will be
+parsed from GitHub as it passes this value on release to the Workflow
+automatically.
+
+Configure ANSIBLE_GALAXY_TOKEN Secret
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
webknjaz, line 207:
```suggestion
Configure ``ANSIBLE_GALAXY_TOKEN`` Secret
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

docs/galaxy_releases.rst
webknjaz: Also, have you checked that ansible that happens to be already pre-installed on the runners is good enough for building&publishing collections? It'd mean that it may be unnecessary to also install it manually.

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+
+.. code-block:: yaml
+
+  name: "release"
+  on:
+  release:
+    types:
+      - created
+  jobs:
+  release:
```
webknjaz, line 177: the indentation here is broken

webknjaz, line 177: Also you should add a name for the job

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+-------------------------
+
+To be able to execute the Ansible playbook we need to configure the GitHub
+Actions Workflow. To do this we will need to create the
+:code:`.github/workflows/` folder if it does not already exist. Then create a
+new file :code:`release.yml` and it should look something like this.
+
+.. code-block:: yaml
+
+  name: "release"
```
webknjaz, line 171: be consistent with the other strings in this snippet
```suggestion
  name: release
```

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,219 @@
...
+  # The namespace of the collection. This can be a company/brand/organization or product namespace under which all
+  # content lives. May only contain alphanumeric lowercase characters and underscores. Namespaces cannot start with
+  # underscores or numbers and cannot contain consecutive underscores
+  namespace: namespace_name
+
+  # The name of the collection. Has the same character restrictions as 'namespace'
+  name: collection_name
+
+  # The version of the collection. Must be compatible with semantic versioning
+  version: "{{ tag }}"
```
webknjaz, line 111: AFAICS this is the only value that's being rendered in this template, it seems to be overengineered and could be solved with a simple `sed`.

docs/galaxy_releases.rst
ericsysmin: So what do you do when 20.04 is the latest and doesn't have the expected packages or support?

docs/galaxy_releases.rst
ericsysmin: No, what I pasted actually works, @geerlingguy used it and I used it that same way

docs/galaxy_releases.rst
ericsysmin: Didn't know about the `::add-path::` value.

docs/galaxy_releases.rst
ericsysmin: make the code suggestion and we can commit it

docs/galaxy_releases.rst
ericsysmin: Because it's not passed as a param to the publish, instead we're reading from .ansible

docs/galaxy_releases.rst
ericsysmin: Awww, it's like we don't even need Ansible for things ;)

docs/galaxy_releases.rst
ericsysmin: it is named, however, tabs were messed up here, fixing it

docs/galaxy_releases.rst
ericsysmin: What are the proposed steps then?

docs/galaxy_releases.rst
ericsysmin: Should we modify the pip upgrade step as well to use --user then?

docs/galaxy_releases.rst
ericsysmin: But if you notice tag isn't provided by the script, it's actually being processed as github.ref includes `/` inside because it's a release ref, not a number.

docs/galaxy_releases.rst
ericsysmin: @felixfontein I don't have many collections where I develop modules themselves rather work on other's modules or adding them for other projects, so I don't have a large collection to test or use. So it's relevant for the purpose of this commit, and this is why we're collaborating on this. If you can make some commits to this PR that would be great.

docs/galaxy_releases.rst
ericsysmin: Assuming latest is best is never a good way to go, at least from any operations perspective because latest is unpredictable in most including docker as a tag of latest can and may get moved to a release your code isn't tested on.

webknjaz: there was a problem when GH was moving fast and dropped support for previous macOS versions, for example. and it was PITA to upgrade the hardcoded thing everywhere. I'm just trying to avoid the same. Besides, it's not like you're using some extraordinary software here that may suddenly disappear.

docs/galaxy_releases.rst
webknjaz: yep, I saw that. Btw you could easily get it from the Git checkout like `setuptools-scm` does this which is even better because this'd be portable and could run from the local machine as well

docs/galaxy_releases.rst
webknjaz: I believe that it works. But my concern here is readability.

docs/galaxy_releases.rst
webknjaz: If you're up to writing a few lines of bash and a dockerfile, you could create an action or two. The question is whether it's overkill for this task, though.

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. code-block:: yaml
+
+  name: release
+  on:
+  release:
+    types:
+      - created
+  jobs:
+    release:
```
webknjaz, line 168:
```suggestion
    release:
      name: Build and release a collection
```

docs/galaxy_releases.rst
webknjaz: No, you only named a yaml key, but there's also `name: ` inside that can be specified for better representation in the UI.

docs/galaxy_releases.rst
webknjaz: It'd install pip into a user dir too. Better question would be — do we really have to upgrade pip?

geerlingguy:
FYI: https://github.com/artis3n/ansible_galaxy_collection

It's a GitHub Action that has support for publishing collections. My main concern with any automation, however, is that right now for any collection where I upload my personal access token, if there's any way someone could sneak in a commit, they would immediately have access to publishing to _any_ other namespace/collection to which I have access—that needs to be an extremely bold and highlighted warning anywhere automation is concerned. See https://github.com/ansible/galaxy/issues/2275 and https://github.com/ansible/galaxy/issues/2276.

docs/galaxy_releases.rst
ericsysmin: I was just using the params that github provides internally, are you able to use github actions locally? or should we just delete this PR and move to using tags and using more than github workflow for releases?

ericsysmin:
@geerlingguy I'm curious if using the token in the command to ansible exposes it on the github action, at least with the ENV var it doesn't expose it to the console I don't believe

docs/galaxy_releases.rst
```diff
@@ -0,0 +1,213 @@
...
+        register: galaxy_yml
+      - name: Build the collection. # noqa 503
+        command: ansible-galaxy collection build
+        args:
+          chdir: ../
+        when: galaxy_yml.changed
+      - name: Publish the collection. # noqa 503
+        command: >
+          ansible-galaxy collection publish ./namespace-collection-{{ tag }}.tar.gz
+            --api-key={{ ansible_galaxy_token }}
```
ericsysmin, line 76: We apparently can't do this because it would put the token in the std out
```n role test48s

Run ansible-playbook -i 'localhost,' build/galaxy_deploy.yml -e "github_tag=refs/tags/1.0.0"

PLAY [localhost] ***************************************************************

TASK [Ensure the ANSIBLE_GALAXY_TOKEN environment variable is set.] ************
skipping: [localhost]

TASK [Ensure the ~/.ansible directory exists.] *********************************
[DEPRECATION WARNING]: Distribution Ubuntu 18.04 on host localhost should use 
/usr/bin/python3, but is using /usr/bin/python for backward compatibility with 
ok: [localhost]```

Comment (ericsysmin) [reply docs/galaxy_releases.rst]: Let me try and see what happens.

**Comment by ericsysmin** [comment]
This is the tested output so far, and removing the apikey from the environment.

```

PLAY [localhost] ***************************************************************

TASK [Ensure that the ansible_galaxy_apikey exists] ****************************
ok: [localhost] => {
    "changed": false,
    "msg": "All assertions passed"
}

TASK [Template out the galaxy.yml file.] ***************************************
changed: [localhost]
[DEPRECATION WARNING]: Distribution Ubuntu 18.04 on host localhost should use 
/usr/bin/python3, but is using /usr/bin/python for backward compatibility with 
prior Ansible releases. A future Ansible release will default to using the 
discovered platform python for this host. See https://docs.ansible.com/ansible/
2.9/reference_appendices/interpreter_discovery.html for more information. This 
feature will be removed in version 2.12. Deprecation warnings can be disabled 
by setting deprecation_warnings=False in ansible.cfg.

TASK [Build the collection.] ***************************************************
changed: [localhost]

TASK [Publish the collection.] *************************************************
changed: [localhost]

PLAY RECAP *********************************************************************
localhost                  : ok=4    changed=3    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

**Comment by artis3n** [comment]
> @geerlingguy I'm curious if using the token in the command to ansible exposes it on the github action, at least with the ENV var it doesn't expose it to the console I don't believe

If you store the token in a repo secret, the output of the value is masked anywhere it is present in the logs: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#about-encrypted-secrets

The risk with the token, as @geerlingguy points out, is that anyone with write access to the repo could push changes to the workflow file and have access to the secret to use in arbitrary bash. Forks do not get access to secrets in the original's repo, but then the workflow does not run successfully from a fork.

**Comment by artis3n** [comment]
FWIW, once something like https://github.com/ansible/galaxy/issues/2276 exists I'd be happy to make supported changes to my Collection action that is mentioned by geerlingguy above.

**Comment by ericsysmin** [comment]
> > @geerlingguy I'm curious if using the token in the command to ansible exposes it on the github action, at least with the ENV var it doesn't expose it to the console I don't believe
> 
> If you store the token in a repo secret, the output of the value is masked anywhere it is present in the logs: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#about-encrypted-secrets
> 
> The risk with the token, as @geerlingguy points out, is that anyone with write access to the repo could push changes to the workflow file and have access to the secret to use in arbitrary bash. Forks do not get access to secrets in the original's repo, but then the workflow does not run successfully from a fork.

That user would also need permissions to create the release as well.

**Comment by ericsysmin** [comment]
That sounds almost like a fundamental flaw if you can't trust the people who manage the repo. They should also likely have access to the secret in the Settings -> Secrets right? Since the secret isn't available outside of the repo that also prevents anyone from getting that secret locally. No one should run this workflow locally. It's an instruction specifically for GitHub Actions Workflow which isn't local. However, another section could be dedicated to automating local release actions.

**Comment by artis3n** [comment]
> They should also likely have access to the secret in the Settings -> Secrets right?

Repo owners with "owner" or "admin" privileges can add or edit the values of secrets, but you can't read the secret values once saved. "Write" contributors could not get to that menu.

> manage the repo. They should also likely have access to the secret in the Settings -> Secrets right? Since the secret isn't available outside of the repo that also prevents anyone from getting that secret locally

True, but if someone has write access they can add a `run` task on the workflow to curl the secret to their own server, or something. They could modify the `on` trigger so the workflow runs on their commit, no PR or release required.

> GitHub doesn't provide any way of preventing the exposure on secrets if the person has write access. 

Yup. Presumably someone with write access to a repo is trusted. I do think geerlingguy's point:

> if there's any way someone could sneak in a commit, they would immediately have access to publishing to any other namespace/collection to which I have access

is important to keep in mind. But, for my collection action (https://github.com/artis3n/ansible_galaxy_collection), I do have users pass their galaxy token via a Secret in their repo. That is the 'right' way to do it, but there is a risk with users with write permission inherent to Workflows that should probably be highlighted if a workflow suggestion is added to the docs.

> That sounds almost like a fundamental flaw if you can't trust the people who manage the repo.

Yeah, totally agree.

**Comment by ericsysmin** [comment]
Ok, so maybe we should add these as details somewhere in the docs?

**Comment by geerlingguy** [comment]
> That sounds almost like a fundamental flaw if you can't trust the people who manage the repo.

Just being the devil's advocate, when it comes to the possibility of any of the dozens of people maintaining the `community` collections which are going to be used across most Ansible installations after 2.10 having any of their personal tokens compromised across any of their repositories (if we have them all use their single personal access token for Galaxy across them all)... I'd rather have a belt-and-suspenders approach. One token that has access to everything in Galaxy is a bit risky.

This is probably not the main place for the security discussion, but we should make sure it is highlighted that if anyone were to get access to your token, they could post to any namespace/collection in Galaxy that you maintain.

**Comment by gundalow** [comment]
For information, and `community.` and `ansible.` collections (which are hosted in https://github.com/ansible-collections/) will use Zuul to publish to Galaxy. 

This means even someone has admin rights on the repo they will not be able to get the Galaxy Token.

**Comment by ericsysmin** [comment]
@gundalow is there a doc on how Zuul is used to publish to Galaxy?

**Comment by pabelanger** [comment]
@ericsysmin we have a release-ansible-collection-galaxy job, that is available to all project which are using zuul for CI. https://dashboard.zuul.ansible.com/t/ansible/build/f3b10c3ed59c4584a081e245a5900dba/console is an example job run. Much like this code, there is a service account on galaxy, which we use the token for, to do the publish.

Then, when ever a project tags a release (or merges a commit) we build and publish to galaxy.

File: docs/galaxy_releases.rst (hunk 13 above)
Comment (webknjaz) [reply docs/galaxy_releases.rst:103]: I said above assuming you're adding this extra ansible playbook layer because you want this automation to be portable and run both locally (w/o  GH actions) and in CI/CD.
If you only intend to have a GHA based solution you don't really need this many levels of indirection.

**Comment by webknjaz** [comment]
@ericsysmin FYI if you want to explicitly hide something from the log, like secrets, you can also issue a workflow command: https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#masking-a-value-in-log

**Comment by webknjaz** [comment]
> FYI there's already a promising GitHub Action for publishing collections by @artis3n here: https://github.com/artis3n/ansible_galaxy_collection

That action also mixes up two stages that are meant to be separate: build + publish. I'm strongly against promoting such an approach.
Ideally, it should be possible to test what's going to be published rather than just source. When build+publish are squashed there's a possibility that you'll be testing not exactly what the user will get. And this flow looks as follows:
1. test the src.
2. build a tarball + immediately publish it w/o testing that tarball.

What I advocate for is the following:
1. Build a tarball and store it as an artifact.
2. Have a test matrix that downloads that tarball, installs it and tests its contents (this is most likely a separate job or a collection of jobs).
3. Have a publish step that downloads the very same tarball that's been tested and uploads it to galaxy. This ensures that this step doesn't have a different build that's being tested.

**Comment by artis3n** [comment]
@webknjaz that's a great point, and that's a flow I'll look into supporting. Feel free to file an issue if you'd like.

**Comment by pabelanger** [comment]
> > FYI there's already a promising GitHub Action for publishing collections by @artis3n here: https://github.com/artis3n/ansible_galaxy_collection
> 
> That action also mixes up two stages that are meant to be separate: build + publish. I'm strongly against promoting such an approach.
> Ideally, it should be possible to test what's going to be published rather than just source. When build+publish are squashed there's a possibility that you'll be testing not exactly what the user will get. And this flow looks as follows:
> 
> 1. test the src.
> 2. build a tarball + immediately publish it w/o testing that tarball.
> 
> What I advocate for is the following:
> 
> 1. Build a tarball and store it as an artifact.
> 2. Have a test matrix that downloads that tarball, installs it and tests its contents (this is most likely a separate job or a collection of jobs).
> 3. Have a publish step that downloads the very same tarball that's been tested and uploads it to galaxy. This ensures that this step doesn't have a different build that's being tested.

This is what we do for testing network / security / content collections for ansible team.  It works very well, and also advocate for people to use.  One addition we are going to add, is an ephemeral galaxy server where we can store the artifact for testing, then promote to galaxy once code is merged.

**Comment by webknjaz** [comment]
> One addition we are going to add, is an ephemeral galaxy server where we can store the artifact for testing, then promote to galaxy once code is merged.

Yeah, ansible/ansible now has "falaxy" for testing. Not sure if it can be used outside, though.



> @webknjaz that's a great point, and that's a flow I'll look into supporting. Feel free to file an issue if you'd like.

I don't have anything else to add there so you can copy what I wrote into an issue as you see fit. I'm currently implementing a similar workflow for a Python package distribution on GitHub Actions Workflows so it's slightly different packaging-wise but the you may get some high-level inspiration ideas from https://github.com/ansible/pylibssh/actions?query=workflow%3A%22%F0%9F%8F%97+%F0%9F%93%A6+%26+test+%26+publish%22 (some bits are still in progress tho).

**Comment by ericsysmin** [comment]
> > FYI there's already a promising GitHub Action for publishing collections by @artis3n here: https://github.com/artis3n/ansible_galaxy_collection
> 
> That action also mixes up two stages that are meant to be separate: build + publish. I'm strongly against promoting such an approach.
> Ideally, it should be possible to test what's going to be published rather than just source. When build+publish are squashed there's a possibility that you'll be testing not exactly what the user will get. And this flow looks as follows:
> 
> 1. test the src.
> 2. build a tarball + immediately publish it w/o testing that tarball.
> 
> What I advocate for is the following:
> 
> 1. Build a tarball and store it as an artifact.
> 2. Have a test matrix that downloads that tarball, installs it and tests its contents (this is most likely a separate job or a collection of jobs).
> 3. Have a publish step that downloads the very same tarball that's been tested and uploads it to galaxy. This ensures that this step doesn't have a different build that's being tested.

Why do you have to be so correct....lol Can you make changes to this PR with that order?

**Comment by webknjaz** [comment]
No, I don't have powers in this repo :)

**Comment by webknjaz** [comment]
@ericsysmin you can find some inspiration on how to work with upload/download artifact @ https://github.com/ansible-community/collection_migration/blob/master/.github/workflows/collection-migration-tests.yml#L136-L188. Except that both now have `v2` released and the download one can fetch all the artifacts, not just one by name.

**Comment by geerlingguy** [comment]
> 1. Build a tarball and store it as an artifact.
> 2. Have a test matrix that downloads that tarball, installs it and tests its contents (this is most likely a separate job or a collection of jobs).
> 3. Have a publish step that downloads the very same tarball that's been tested and uploads it to galaxy. This ensures that this step doesn't have a different build that's being tested.

I have a feeling we're talking about vastly different use cases at this point. For a certified content collection with Ansible modules that are organized by an enterprise vendor with a team of developers able to support it, having tooling in place for multiple layers of build, test, and deploy processes may be a worthy goal.

But I think both @ericsysmin and myself are coming at this from the approach of long-time Galaxy role authors.

We are used to having a lightweight CI system, a lightweight release system, and a lightweight package manager. We're coming from the angle of someone more used to systems like NPM, Packagist, or Rubygems, not from the angle of building and distributing RPMs, .debs, etc.

I think theres a major conflict in goals between the two use cases for collections. I just want there to be some easy way for me to maintain collections like I did roles, especially if [roles are going to vanish in Galaxy next-gen](https://github.com/ansible/galaxy_ng/issues/58)... because if it's an onerous process (like it is today), I know I'm not going to do it.

I don't want to have to tie two or three different backend systems/processes into each collection I maintain, and I'm probably more willing to bend on this topic than most people who have contributed roles in the past...

**Comment by ericsysmin** [comment]
@geerlingguy it seems maybe a middle ground is that we do a build first, then do the install instead of doing this
```
before_script:
  - cd ../
  - mkdir -p ansible_collections/$COLLECTION_NAMESPACE
  - mv ansible-collection-$COLLECTION_NAME ansible_collections/$COLLECTION_NAMESPACE/$COLLECTION_NAME
  - cd ansible_collections/$COLLECTION_NAMESPACE/$COLLECTION_NAME
```

**Comment by ericsysmin** [comment]
so really we'd be doing
```
before_script:
  - ansible-galaxy collection build
  - ansible-galaxy collection install my_namespace-my_collection-1.0.0.tar.gz

webknjaz:
@geerlingguy yeah, I was this approach a lot in the past and even now in the Python ecosystem, with publishing to PyPI and have it simplified like this in a lot of places.
One reason for this was that the public CIs didn't have a mature artifact and stage pipelining support. I feel like with GHA it became so much easier enabling virtually anyone can allow themselves a more sophisticated approach, especially if there'd be some pre-baked template that anybody could use.
Of course, some people may still want to opt out. And that's okay for them. OTOH we should show a more complete approach. Folks can then just remove the steps they dislike.

pabelanger:
> so really we'd be doing
> 
> ```
> before_script:
>   - ansible-galaxy collection build
>   - ansible-galaxy collection install my_namespace-my_collection-1.0.0.tar.gz
> ```

Yes, this is much better real world testing, then manually moving, creating the install manually.

ericsysmin:
SOoooo, trying this out, it may not be as easy as hoped, it also will require galaxy.yml being present at all times, so no version generation allowed, so people will have to always modify that file on release.

gundalow:
I see https://github.com/ansible-community/molecule/pull/2690 has been merged.

What do people think the way forward is here?

ssbarnea:
One thing that I find bit mind-blowing is that there is no command `ansible-galaxy collection ls`. No official way to list what collections are installed.

Also there is no `rm` either and that does not help much with automation or debugging.

felixfontein:
It's called `ansible-galaxy collection list` and has been working in devel since for some months now. :) (No `rm`/`uninstall` though.)

pabelanger:
> One thing that I find bit mind-blowing is that there is no command `ansible-galaxy collection ls`. No official way to list what collections are installed.
> 
> Also there is no `rm` either and that does not help much with automation or debugging.

Created https://github.com/ansible/ansible/issues/60292 a while back, happy to say code was added.

ssbarnea:
While I do have jobs testing deve, I am not brave enough to run devel locally so I didn't know about the list feature. Probably it should have being ported to latest stable.

Andersson007:
@acozine SGTM, closing, thanks everyone!
If anyone is interested to proceed, feel free to re-open (or ping me)

PR #90, opened 2020-07-09
Collection checklist

gundalow:
Requirements for Collections to be included in Ansible

This is currently in bullet point form, we can expand over time.

This could live elsewhere, though I realise it hasn't been captured at all so far.

Commits to update wording here are welcome.

PR #94, opened 2020-07-29
Add details on versioning and deprecation

felixfontein:
Updates #90 with information on versioning and deprecation.

Should be redirected to master once #90 is merged; I'm keeping it directed at #90's branch for now so it's easier to see the differences.

CC @gundalow @abadger @acozine @samccann

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -1,0 +1,114 @@
+*****************************
+Ansible Collections Checklist
+*****************************
+
+.. contents:: Topics
+
+Overview
+========
+
+**Warning:** Subject to frequent updates
+       This is a "living document", expect it to change as we progress with the Collections work over the next few months.
+
+Feedback
+========
+
+As with any project it's very important that we get feedback from users, contributors and maintainers. We recognize that the move to Collections is a big change in how Ansible is developed and delivered.
+
+Please raise feedback by
+
+* Discussing in ``#ansible-community`` in Freenode IRC
+* Adding to the `Community Working Group IRC meeting <https://github.com/ansible/community/issues/539>`_
+* Creating `GitHub Issues <https://github.com/ansible-collections/overview/issues>`_ against this repo
+
+
+Why is this needed
+===================
+
+
+Repo structure
+===============
+
+galaxy.yml
+----------
+
+* tags MUST set
+* dependencies MUST be set to ``'>=1.0.0'`` (Might have to think about dependency order when publishing)
+
+meta/runtime.yml
+----------------
+Example: `meta/runtime.yml <https://github.com/ansible-collections/collection_template/blob/main/meta/runtime.yml>`_
+* MUST define the minimim version of Ansible which this collection works with
+  * If the Collection works with Ansible 2.9, then this shuld be set to
+
+Modules & Plugins
+------------------
+
+Documentation
+~~~~~~~~~~~~~~
+All module and plugin ``DOCUMENTATION`` MUST
+
+* Use the FQCN for ``M(...)`` FIXME Docs link
+* Use the FQCN for ``SEEALSO`` FIXME Docs link
+* Use the FQCN for ``extends_documentation_fragment:``, unles you are referring to doc_fragments from ansible-base
+
+Contributor Workflow
+====================
+
+Changelogs
+~~~~~~~~~~
+
+FIXME: What's the actual minimum requirement here, is it a file called X?
+
+FIXME: Add details and examples of automated system
+
+Versioning and deprecation
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+FIXME Details here
+* Deprecation vs Collection version
+* How many Collection major versions to give
+
+Repo management
+===============
+
+Repo name
+~~~~~~~~~
+
+For collections under ansible-collections the repo SHOULD be named ``NAMESPACE.COLLECTION``
+
+Branch name and configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+All new repos under `ansible-collections<https://github.com/ansible-collections>`_ MUST have `main`` as the default branch.
+
+Existing repos SHOULD be converted to use ``main``
+
+Repo Protections:
+* Allow merge commits: disallowed
+
+Branch protections MUST be enforced
+* Require linear history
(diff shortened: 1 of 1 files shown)


collection_requirements.rst
```diff
@@ -87,9 +87,15 @@ Please note that the porting guide is compiled from ``changelogs/changelog.yaml`
 Versioning and deprecation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-FIXME Details here
-* Deprecation vs Collection version
-* How many Collection major versions to give
+* Every ansible minor version series (2.10.x) will keep the major version of a collection constant. So if ``community.general`` 1.2.0 is included in 2.10.0, 2.10.x will contain the latest ``community.general`` 1.y.z release that was available when building 2.10.x. If ``community.general`` 2.0.0 is released, it will **not** be included in ansible 2.10.x, but only in 2.11.0 (if it was published before that was released).
```
acozine, line 90:
```suggestion
* To preserve backward compatibility for users, every ansible minor version series (2.10.x) will keep the major version of a collection constant. If ansible 2.10.0 includes ``community.general`` 1.2.0, then each 2.10.x release will include the latest ``community.general`` 1.y.z release available at build time. Ansible 2.10.x will never include a ``community.general`` 2.y.x release, even if it is available. Major collection version changes will be included in the next ansible minor release (2.11.0, 2.12.0, and so on).
```

collection_requirements.rst
```diff
@@ -87,9 +87,15 @@ Please note that the porting guide is compiled from ``changelogs/changelog.yaml`
...
+* Therefore, please make sure that the current major release of your collection included in 2.10.0 receives at least bugfixes as long new 2.10.x releases are produced.
+* Since new minor releases are included, you can include new features, modules and plugins. You must make sure that you do not break backwards compatibility! (See `semantic versioning <https://semver.org/>`_.) This means in particular:
+  * You can fix bugs in patch releases, but not add new features or deprecate things.
+  * You can add new features and deprecate things in minor releases, but not remove things or change behavior of existing features.
+  * Removals and breaking changes can only be made in major releases.
```
acozine, line 95:
```suggestion
  * You can only remove things or make breaking changes in major releases.
```

felixfontein:
Removed the [WIP] since the original PR has been merged.

collection_requirements.rst
```diff
@@ -87,9 +87,15 @@ Please note that the porting guide is compiled from ``changelogs/changelog.yaml`
 Versioning and deprecation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-FIXME Details here
-* Deprecation vs Collection version
-* How many Collection major versions to give
+* To preserve backward compatibility for users, every ansible minor version series (2.10.x) will keep the major version of a collection constant. If ansible 2.10.0 includes ``community.general`` 1.2.0, then each 2.10.x release will include the latest ``community.general`` 1.y.z release available at build time. Ansible 2.10.x will never include a ``community.general`` 2.y.x release, even if it is available. Major collection version changes will be included in the next ansible minor release (2.11.0, 2.12.0, and so on).
+* Therefore, please make sure that the current major release of your collection included in 2.10.0 receives at least bugfixes as long new 2.10.x releases are produced.
+* Since new minor releases are included, you can include new features, modules and plugins. You must make sure that you do not break backwards compatibility! (See `semantic versioning <https://semver.org/>`_.) This means in particular:
+  * You can fix bugs in patch releases, but not add new features or deprecate things.
```
gundalow, line 93: I think you need a blank line when changing bullet-list indent levels
```suggestion

  * You can fix bugs in patch releases, but not add new features or deprecate things.
```

Andersson007:
1. We could also add to the check list a strong recommendation to create a pinned issue about "Releasing, Versioning and Deprecation" describing the process in detail in every collection repository and put here a reference to https://github.com/ansible-collections/community.general/issues/582 as an example.
@felixfontein if you're agree you could add it right now

2. Do i understand correct that, according to the changes, if a major release of collection is released right after 2.10 (or any 2.X), people will need to wait for ~4-6 month until the next Ansible 2.X is released, correct?
The idea is sensible but the possible term of waiting doesn't feel good:)
So, would be good to inform users about this in some time before 2.X release, e.g. by creating an issue (as a reminder) in every repo.
It would give maintainers some time to release major versions if needed.
We could do it by a script as i suggested once, i have one to copy issues from one repo to another, i could run it in a loop via a file containing all links.

felixfontein:
> 1. We could also add to the check list a strong recommendation to create a pinned issue about "Releasing, Versioning and Deprecation" describing the process in detail in every collection repository and put here a reference to [ansible-collections/community.general#582](https://github.com/ansible-collections/community.general/issues/582) as an example.
> @felixfontein if you're agree you could add it right now

I added an item for that.

> 2. Do i understand correct that, according to the changes, if a major release of collection is released right after 2.10 (or any 2.X), people will need to wait for ~4-6 month until the next Ansible 2.X is released, correct?

Yes, that's correct. Major versions can contain breaking changes, so we cannot include it.

> The idea is sensible but the possible term of waiting doesn't feel good:)
> So, would be good to inform users about this in some time before 2.X release, e.g. by creating an issue (as a reminder) in every repo.

I think we should try to avoid this "create issue in every repo" thing, but instead use ansible-collections/overview#45 and/or The Bullhorn to announce such things. Otherwise we have a lot of extra work :)

Also a problem is that not every collection repo is managed on GitHub.

> We could do it by a script as i suggested once, i have one to copy issues from one repo to another, i could run it in a loop via a file containing all links.

Creating the issue is (at least when doing it once) pretty simple, the hardest part is first finding the collection repos. So if we only do this once, doing it manual is probably not that bad.

Andersson007:
```
I think we should try to avoid this "create issue in every repo" thing, but instead use #45 and/or The Bullhorn to announce such things. Otherwise we have a lot of extra work :)

Also a problem is that not every collection repo is managed on GitHub.
```
Agreed. I asked because, as far as i remember, we were going to create the issue. Using 45th and the newsletter will make lives a bit easier:)

Andersson007:
```I added an item for that.```
LGTM, thanks

felixfontein:
Thanks everyone for reviewing and helping improving this!

PR #103, opened 2020-08-06
collection_requirements.rst: add naming recommendations

Andersson007:
##### SUMMARY
collection_requirements.rst: add naming recommendations

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -110,7 +110,13 @@ Repository management
 Repository name
 ---------------
 
-For collections under ansible-collections the repo SHOULD be named ``NAMESPACE.COLLECTION``
+For collections under ansible-collections the repository SHOULD be named ``NAMESPACE.COLLECTION``.
+
+For collections created for working with a particular entity, they should contain the entity name, for example ``community.mysql``.
+
+We should avoid:
+* too long FQCN names: try to make it compact but clear
+* nested namespaces if this is not really necessary, especially namespaces with repetitive parts
 
 Branch name and configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


collection_requirements.rst
```diff
@@ -110,7 +110,13 @@ Repository management
 Repository name
```
bmillemathias, line 110: s/Repository name/Name Conventions/

collection_requirements.rst
```diff
@@ -110,7 +110,13 @@ Repository management
...
 ---------------
 
-For collections under ansible-collections the repo SHOULD be named ``NAMESPACE.COLLECTION``
+For collections under ansible-collections the repository SHOULD be named ``NAMESPACE.COLLECTION``.
```
bmillemathias, line 113: under ansible-collections -> from ansible-collections github project ?
Should be we talk about convention for corporate maintained collection (`COMPANY_NAME.PRODUCT` or `COMPANY_NAME.SERIES`)

collection_requirements.rst
```diff
@@ -110,7 +110,13 @@ Repository management
...
+
+For collections created for working with a particular entity, they should contain the entity name, for example ``community.mysql``.
+
+We should avoid:
+* too long FQCN names: try to make it compact but clear
+* nested namespaces if this is not really necessary, especially namespaces with repetitive parts
```
bmillemathias, line 119: I don't understand what is nested namespaces

bmillemathias: having some do and don't example would be good.

Andersson007: i think something like awx.awx.awx

Andersson007: if it's technically possible, i'm not 100% sure

Andersson007: i'll think of how to reformulate that

bmillemathias: Not an easy task to formulate all this. we know
* we would like to avoid having the same word for `NAMESPACE` and `COLLECTION` (like `awx.awx` or `gluster.gluster` however I think we should not give the name of project that already exists). We need also to give advice for good naming but this is really difficult for collections dedicated on a specific product unless there is a company used as namespace.
* For the modules/plugins/... we can say that collections dedicated to a product should not use the product in each object name

Andersson007: @bmillemathias thanks for the recommendations, could you look at https://github.com/ansible-collections/overview/pull/105
I partly added what you suggested.

PR #133, opened 2020-12-09
BSD License for module_utils should be 2 clause

abadger:
The current ansible module_utils are actually under the BSD-2-clause license, not the BSD-3-clause license.

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


abadger:
Also change the short name for GPLv3+ to the SPDX identifier and link to the text of both licenses.  But the version change is the only thing that changes the meaning of the text.

PR #134, opened 2020-12-09
MUST be published to Galaxy + README.md

gundalow:
##### SUMMARY

* Collections MUST be on Galaxy
* Collections MUST have a `README.md`

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -38,6 +38,7 @@ Collection Infrastructure
 
 * MUST have a publicly available issue tracker, that does not require a paid level of service to create an account or view issues.
 * MUST use the `Ansible Code of Conduct (CoC) <https://docs.ansible.com/ansible/latest/community/code_of_conduct.html>`_ by copying `CODE_OF_CONDUCT.md <https://raw.githubusercontent.com/ansible-collections/.github/main/CODE_OF_CONDUCT.md>`_ into the root of your Git repository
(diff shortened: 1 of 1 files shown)


collection_requirements.rst
```diff
@@ -38,6 +38,7 @@ Collection Infrastructure
 
 * MUST have a publicly available issue tracker, that does not require a paid level of service to create an account or view issues.
 * MUST use the `Ansible Code of Conduct (CoC) <https://docs.ansible.com/ansible/latest/community/code_of_conduct.html>`_ by copying `CODE_OF_CONDUCT.md <https://raw.githubusercontent.com/ansible-collections/.github/main/CODE_OF_CONDUCT.md>`_ into the root of your Git repository
+* MUST be published to `Ansible Galaxy <https://galaxy.ansible.com>_.
```
dmsimard, line 41: Should this be here or towards the end in "Requirements for collections to be included in the Ansible Package" or both ?

felixfontein: I think it is fine here, since it should also be a requirement for any serious collection that's not for internal use.

abadger:
This content was approved in today's IRC meeting.

dmsimard (approved):

collection_requirements.rst
gundalow: 2020-12-09 IRC meeting we said we were OK with this as is

PR #132, opened 2020-12-02
Must change changes to the checklist criteria

abadger:
Update CI testing that:
* Sanity tests must pass
* Initial rules around use of ignores.txt

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

PR #137, opened 2020-12-09
Need approval of the CI changes

abadger:
The substantive changes in
https://github.com/ansible-collections/overview/pull/132/commits/b4c59b2f96338dab54e7bef0f400d678f9fc12c4
were not approved in the meeting.

However, the checklist as a whole was approved in the next meeting.

This leaves the text in this PR in an odd situation where it might be
approved or might not.  This PR is not necessarily the proposal; tather
the change needs to be discussed, possibly a new proposal needs to be
created, and it needs to be voted approved at the next meeting.

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -182,6 +182,8 @@ CI Testing
   * All entries in ignores.txt MUST have a justification in a comment in the ignore.txt file for each entry.  For example ``plugins/modules/docker_container.py use-argspec-type-path # uses colon-separated paths, can't use type=path``.
   * Reviewers can block acceptance of a new collection if they don't agree with the ignores.txt entries.
 
+* You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc)  of``ansible-base``/``ansible-core`` that the collection supports. (Usually the head of the stable-xxx branches.)
+
 * All CI tests MUST run regularly (nightly, or at least once per week) to ensure that repos without regular commits are tested against the latest version of ansible-test from each ansible-base/ansible-core version tested. 
 
 All of the above can be achieved by using the following GitHub Action template, see `example <https://github.com/ansible-collections/collection_template/tree/main/.github/workflows>`_.
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -184,6 +184,9 @@ CI Testing
 
 * You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc) of ``ansible-base``/``ansible-core`` that the collection supports. (Usually the ``HEAD`` of the stable-xxx branches.)
 
+* All CI tests MUST run against every pull request and SHOULD pass before merge.
+* All CI tests MUST pass for the commit that releases the collection.
+ 
 * All CI tests MUST run regularly (nightly, or at least once per week) to ensure that repos without regular commits are tested against the latest version of ansible-test from each ansible-base/ansible-core version tested. 
 
 All of the above can be achieved by using the following GitHub Action template, see `example <https://github.com/ansible-collections/collection_template/tree/main/.github/workflows>`_.
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -182,7 +182,7 @@ CI Testing
   * All entries in ignores.txt MUST have a justification in a comment in the ignore.txt file for each entry.  For example ``plugins/modules/docker_container.py use-argspec-type-path # uses colon-separated paths, can't use type=path``.
   * Reviewers can block acceptance of a new collection if they don't agree with the ignores.txt entries.
 
-* You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc)  of``ansible-base``/``ansible-core`` that the collection supports. (Usually the head of the stable-xxx branches.)
+* You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc) of ``ansible-base``/``ansible-core`` that the collection supports. (Usually the ``HEAD`` of the stable-xxx branches.)
 
 * All CI tests MUST run regularly (nightly, or at least once per week) to ensure that repos without regular commits are tested against the latest version of ansible-test from each ansible-base/ansible-core version tested. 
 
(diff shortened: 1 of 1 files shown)


felixfontein:
I think the most important problem with the current version is (as pointed out by @tadeboro) that intermediate commits might not always pass CI, if several are pushed at once (or the collection doesn't use squash+merge). We should better say that CI must pass for the commit that releases the collection.

collection_requirements.rst
```diff
@@ -175,15 +175,13 @@ CI Testing
...
 * You MUST run ``ansible-test sanity`` from the `latest stable ansible-base/ansible-core branch <https://github.com/ansible/ansible/branches/all?query=stable->`_. 
-* You MUST run CI against all versions of ``ansible-base``/``ansible-core`` that the collection supports.
 * You SHOULD suggest to *additionally* run ``ansible-test sanity`` from the ansible/ansible ``devel`` branch so that you find out about new linting requirements earlier.
 * The sanity tests MUST pass.
 
   * Adding some entries to the ``test/sanity/ignore*.txt`` file is an allowed method of getting them to pass. 
   * All entries in ignores.txt MUST have a justification in a comment in the ignore.txt file for each entry.  For example ``plugins/modules/docker_container.py use-argspec-type-path # uses colon-separated paths, can't use type=path``.
   * Reviewers can block acceptance of a new collection if they don't agree with the ignores.txt entries.
 
-* All CI tests MUST run against every PR & commit to the repo.
```
gundalow, line 186: MUST run, SHOULD pass. We understand that occasionally things outside of your control may break your tests.

collection_requirements.rst
```diff
@@ -175,15 +175,13 @@ CI Testing
 ===========
 
 * You MUST run ``ansible-test sanity`` from the `latest stable ansible-base/ansible-core branch <https://github.com/ansible/ansible/branches/all?query=stable->`_. 
-* You MUST run CI against all versions of ``ansible-base``/``ansible-core`` that the collection supports.
```
gundalow, line 178: I vote that this stays

collection_requirements.rst
felixfontein: Is there a way to achieve this with GitHub Actions or AZP? If you push several commits at once, right now (at least for GHA and Shippable) tests are only run for the last commit, not for the ones before.

collection_requirements.rst
abadger: As long as the example github action does this, I vote it stays too.

collection_requirements.rst
abadger: tadeboro also brought up that this could be an undue burden to maintainers.  For instance, a company that pays for every CI run and does not squash their commits could end up paying a lot to do this.  We don't really care about the intermediate commits when we think about ci being run.  Is there another way we could phrase this that would capture what we really want?  Like "CI must be run on the head of every PR and the repository branches that you release from"?

We should probably add the note about SHOULD pass with the explanation you give as well.  That seems like an important exception to mention.

gundalow: Ah, I hadn't considered that.
OK to change to SHOULD

collection_requirements.rst
gundalow: It does

collection_requirements.rst
Andersson007:
```
* All CI tests MUST run against every pull request (last pushed commit to the repository) and MUST pass before merge with the exception when something outside of maintainers control breaks them.
* All CI tests MUST pass for the commit that releases the collection.
```
? or something like that

gundalow: @abadger Hows does the above suggestion sound to you?

collection_requirements.rst
abadger: Approved at today's meeting:  You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc)  of``ansible-base``/``ansible-core`` that the collection supports. (Usually the head of the stable-xxx branches.)

collection_requirements.rst
```diff
@@ -175,15 +175,15 @@ CI Testing
...
 * You SHOULD suggest to *additionally* run ``ansible-test sanity`` from the ansible/ansible ``devel`` branch so that you find out about new linting requirements earlier.
 * The sanity tests MUST pass.
 
   * Adding some entries to the ``test/sanity/ignore*.txt`` file is an allowed method of getting them to pass. 
   * All entries in ignores.txt MUST have a justification in a comment in the ignore.txt file for each entry.  For example ``plugins/modules/docker_container.py use-argspec-type-path # uses colon-separated paths, can't use type=path``.
   * Reviewers can block acceptance of a new collection if they don't agree with the ignores.txt entries.
 
-* All CI tests MUST run against every PR & commit to the repo.
+* You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc)  of``ansible-base``/``ansible-core`` that the collection supports. (Usually the head of the stable-xxx branches.)
```
felixfontein, line 185:
```suggestion
* You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc) of ``ansible-base``/``ansible-core`` that the collection supports. (Usually the head of the stable-xxx branches.)
```

collection_requirements.rst
felixfontein: @abadger ^ the approved version isn't part of this PR

felixfontein: Ah wait, it ended up later in the document. Sorry for the noise!

collection_requirements.rst
Andersson007: @abadger ^ ? If it's ok, i could add

felixfontein: @Andersson007 it's already here (in slightly modified form): https://github.com/ansible-collections/overview/pull/137/files#diff-dacab21f27a6aa218b8dcaafcf6f5685df0ab52f231afa16e7a9ab5284b72c68R200-R201

Andersson007: @felixfontein you were 4 minutes faster than me:) I've just found that in the meeting summary

PR #149, opened 2021-01-22
Add a requirement that plugins must be one of the recognized types

abadger:
Write up of the plugin type collection criteria which was approved at the Community IRC Meeting of 2021-01-20.

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -86,7 +86,7 @@ Example: `meta/runtime.yml <https://github.com/ansible-collections/collection_te
 Modules & Plugins
 ------------------
 
-* Collections MUST only put plugin types recognized by ansible-core into the ``plugins/`` directory at this time.  The recognized plugin types are listed on https://docs.ansible.com/ansible/devel/plugins/plugins.html plus ``terminal``, ``modules``, ``doc_fragments`` and ``module_utils``.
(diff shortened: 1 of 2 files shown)


collection_requirements.rst
```diff
@@ -86,6 +86,16 @@ Example: `meta/runtime.yml <https://github.com/ansible-collections/collection_te
 Modules & Plugins
 ------------------
 
+* Collections MUST only put plugin types recognized by ansible-core into the :file:`plugins/` directory at this time.  The recognized plugin types are listed on https://docs.ansible.com/ansible/devel/plugins/plugins.html plus ``terminal``, ``modules``, and ``module_utils``.
```
felixfontein, line 89: The GitHub RST renderer does not know `:file:`:
```suggestion
* Collections MUST only put plugin types recognized by ansible-core into the ``plugins/`` directory at this time.  The recognized plugin types are listed on https://docs.ansible.com/ansible/devel/plugins/plugins.html plus ``terminal``, ``modules``, ``doc_fragments`` and ``module_utils``.
```
Maybe we should also link to https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/loader.py and say that if something is supported there, but not included in the docs link and in our explicit list, it is also supported.

collection_requirements.rst
```diff
@@ -86,6 +86,16 @@ Example: `meta/runtime.yml <https://github.com/ansible-collections/collection_te
...
+
+  * The following collections have a temporary exception to use the ``plugin_utils``, ``cli_parsers``, ``fact_diff``, and ``validate`` directories for additional plugins during the 2.10 and 3.0 release cycles.  We will figure out a final policy which these collections will need to comply with before ansible-4.0:
```
felixfontein, line 91: 3.0 does not cover 3.1, 3.2, ... :)
```suggestion
  * The following collections have a temporary exception to use the ``plugin_utils``, ``cli_parsers``, ``fact_diff``, and ``validate`` directories for additional plugins during the 2.10 and 3 release cycles.  We will figure out a final policy which these collections will need to comply with before ansible-4:
```

PR #164, opened 2021-04-07
Expand the licenses allowed in collections

abadger:
Allow modules and module_utils to use licenses that are compatible with
the GPLv3+ rather than restricting them to only the GPLv3+.

This is to address the issue raised here: https://github.com/ansible-community/community-topics/issues/2

(Allowing a collection with modules licensed under Apache v2)

##### ISSUE TYPE
t
- Docs Pull Request

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -220,25 +220,30 @@ Licensing
 **Note**: The guidelines below are more restrictive than strictly necessary.  We will try to add
 a larger list of acceptable licenses once we have approval from Red Hat Legal.
 
-There are three types of plugin code in collections which licensing has to address in different
+There are four types of content in collections which licensing has to address in different
 ways:
 
-:modules: must be licensed compatibly with the `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_
-:module_utils: must be licensed compatibly with the `GPL-3.0-or-later
-               <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  Ansible itself typically
-               uses the `BSD-2-clause <https://opensource.org/licenses/BSD-2-Clause>`_ license to
-               make it possible for third-party modules which are licensed incompatibly with the
-               GPLv3 to use them.  Please consider this use case when licensing your own
-               ``module_utils``.
-:All other plugin code: All other code must be under the `GPL-3.0-or-later
+:modules: must be licensed with a free software license that is compatible with the
+          `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_
+:module_utils: must be licensed with a free software license that is compatible with the
+               `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  Ansible
+               itself typically uses the `BSD-2-clause
+               <https://opensource.org/licenses/BSD-2-Clause>`_ license to make it possible for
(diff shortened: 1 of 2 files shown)


collection_requirements.rst
```diff
@@ -217,7 +217,30 @@ We should avoid FQCN / repository names:
...
-At the moment, ``module_utils`` must be licensed under the `BSD-2-clause <https://opensource.org/licenses/BSD-2-Clause>`_ or `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_ license and all other content must be licensed under the `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  We will have a list of other open source licenses which are allowed as soon as we get Red Hat's legal team to approve such a list for us.
+**Note**: The guidelines below are more restrictive than strictly necessary.  We will try to add
+a larger list of acceptable licenses once we have approval from Red Hat Legal.
+
+There are three types of plugin code in collections which licensing has to address in different
+ways:
+
+:modules: must be licensed compatibly with the `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_
+:module_utils: must be licensed compatibly with the `GPL-3.0-or-later
+               <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  Ansible itself typically
```
tadeboro, line 228: I there something missing in this sentence? I would expect something like "Ansible itself typically **uses** ...", but I am not a native speaker.

abadger: Yes, thanks for catching that!

tadeboro (approved):

felixfontein:
I think we should also require something on the license of other files included in the collections, that are not code. I'm not sure what exactly that requirement would be, though.

abadger:
> I think we should also require something on the license of other files included in the collections, that are not code. I'm not sure what exactly that requirement would be, though.

Yeah, I agree but it's an area that I haven't learned a whole lot about in the past so I hesitated to put something in for this update.

felixfontein:
How about also requiring "compatible to GPLv3+"? If there's need we can easily update it (and allow more licenses).

felixfontein:
Oh, should we also say something special about doc_fragments?

felixfontein:
Decided in today's community meeting (+6, -1, 1x0): to be merged as-is.

The general consensus is to be more flexible, but not without RedHat legal giving a green light first.

ssbarnea:
I do think we should also include a list of OSI approved licenses that we consider compatible or we risk into confusing people that do know differences between most licenses and I do not want for make them believe we force them towards using GPL.

Considering that Galaxy already have a requirements to use SPDX licenses, I think we must follow suit and list the subset that is considered ok. Once we have this list I can also help people pick right compatible licenses by including a rule inside the linter.

As I think it may take a while to get a more complete list, we should at least clearly address answer for the most popular ones: Apache, MIT, BSD, GPL and LGPL.

Links:
* https://opensource.org/licenses/alphabetical
* https://spdx.org/licenses/
* https://en.wikipedia.org/wiki/Permissive_software_license

felixfontein:
@ssbarnea if you want to discuss it, it's better to create a new issue in https://github.com/ansible-community/community-topics/issues. A merged PR isn't the best place to discuss such things :)

ssbarnea:
I added https://github.com/ansible-collections/overview/issues/166 so we can track it there.

felixfontein:
@ssbarnea can you please create an issue in the correct repository (I linked above)?

PR #206, opened 2022-06-08
Change requirements related to running tests against release commit

Andersson007:
##### SUMMARY
Relates to https://github.com/ansible-collections/ansible-inclusion/discussions/44#discussioncomment-2889787.
As integration/unit tests are optional anyway and it can be expensive to run them against release commits, i suggest making their run optional.

### DIFF
--- a/collection_checklist.md
+++ b/collection_checklist.md
@@ -43,5 +43,5 @@ Every comment should say whether the reviewer expects it to be addressed, or whe
 - [ ] if `test/sanity/ignore*.txt` exists, it does not contain error codes listed [here](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#ci-testing)
 - [ ] has CI tests up and running against each of the "major versions" of `ansible-base`/`ansible-core` that the collection supports
 - [ ] all CI tests run against every pull request
-- [ ] all CI tests run against a commit that releases the collection; if they don't pass, the collection won't be released
 - [ ] all CI tests run regularly (nightly, or at least once per week)
+- [ ] Sanity tests run against a commit that releases the collection; if they don't pass, the collection won't be released
--- a/collection_checklist.md
+++ b/collection_checklist.md
@@ -40,8 +40,8 @@ Every comment should say whether the reviewer expects it to be addressed, or whe
 **Tests:**
 
 - [ ] passed `ansible-test sanity`
-- [ ] if `test/sanity/ignore*.txt` exists, it does not contain error codes listed [here](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#ci-testing)
+- [ ] if `test/sanity/ignore*.txt` exists, it MUST not contain error codes listed [here](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#ci-testing)
 - [ ] has CI tests up and running against each of the "major versions" of `ansible-base`/`ansible-core` that the collection supports
-- [ ] all CI tests run against every pull request
-- [ ] all CI tests run regularly (nightly, or at least once per week)
-- [ ] Sanity tests run against a commit that releases the collection; if they don't pass, the collection won't be released
+- [ ] all CI tests MUST run against every pull request
+- [ ] all CI tests MUST run regularly (nightly, or at least once per week)
+- [ ] sanity tests MUST run against a commit that releases the collection; if they don't pass, the collection won't be released
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -338,7 +338,7 @@ CI Testing
 
 * You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc) of ``ansible-base``/``ansible-core`` that the collection supports. (Usually the ``HEAD`` of the stable-xxx branches.)
 * All CI tests MUST run against every pull request and SHOULD pass before merge.
-* At least sanity tests MUST run against a commit that releases the collection (if the collection has integration/unit tests, they SHOULD run too); if they don't pass, the collection will NOT be released.
+* At least sanity tests MUST run against a commit that releases the collection; if they do not pass, the collection will NOT be released. If the collection has integration/unit tests, they SHOULD run too; if they do not pass, the errors SHOULD be analyzed to decide whether they should block the release or not.
 * All CI tests MUST run regularly (nightly, or at least once per week) to ensure that repositories without regular commits are tested against the latest version of ansible-test from each ansible-base/ansible-core version tested. The results from the regular CI runs MUST be checked regularly.
 
 All of the above can be achieved by using the `GitHub Action template <https://github.com/ansible-collections/collection_template/tree/main/.github/workflows>`_.
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -49,7 +49,7 @@ Collections
 * MUST be published to `Ansible Galaxy <https://galaxy.ansible.com>`_.
 * SHOULD NOT contain any large objects (binaries) comparatively to the current Galaxy tarball size limit of 20 MB like, for example, package installers for testing purposes.
 * SHOULD NOT contain any unnecessary files like, for example, temporary files created by development tools.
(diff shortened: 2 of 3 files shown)


Andersson007:
any thoughts folks? If there are no objections, i'll create a community topic to vote on this to get merged

mariolenz:
I'm not really sure about this. On the one hand, you want to make integration/unit tests optional for release commits because they can be expensive. On the other hand, I see the requirement to run _all_ CI tests against _every_ pull request and to run _all_ CI tests _regularly_.

I agree that integration/unit tests can be expensive. But if the requierment is to run those (expansive) tests both against every pull request and regulary, I don't see that we save much.

Don't get me wrong, I'm not against relaxing the test requirements. But shouldn't a _release_ commit be better tested than a "normal" PR?

Andersson007:
> I'm not really sure about this. On the one hand, you want to make integration/unit tests optional for release commits because they can be expensive. On the other hand, I see the requirement to run _all_ CI tests against _every_ pull request and to run _all_ CI tests _regularly_.
> 
> I agree that integration/unit tests can be expensive. But if the requierment is to run those (expansive) tests both against every pull request and regulary, I don't see that we save much.
> 
> Don't get me wrong, I'm not against relaxing the test requirements. But shouldn't a _release_ commit be better tested than a "normal" PR?

@mariolenz , all sounds sensible to me.
On the other hand, integration and unit tests are optional for the collection anyway.
imo it feels a bit strange if we, as now,
- in one statement of the requirements make only sanity tests mandatory
- in the other statement require to run all available tests against release commits

This can lead to situations, like in @resmo 's case, when the requirements will force him to remove the integration and unit tests to formally satisfy the requirements.
I would recommend running all the tests but would keep only sanity run mandatory as their presence in collections is mandatory.

collection_requirements.rst
```diff
@@ -49,7 +49,7 @@ Collections
 * MUST be published to `Ansible Galaxy <https://galaxy.ansible.com>`_.
 * SHOULD NOT contain any large objects (binaries) comparatively to the current Galaxy tarball size limit of 20 MB like, for example, package installers for testing purposes.
 * SHOULD NOT contain any unnecessary files like, for example, temporary files created by development tools.
-* MUST only contain objects that follow the :ref:`licensing rules<Licensing>`.
+* MUST only contain objects that follow the :ref:`licensing rules<Licensing>`_.
```
resmo, line 52: is this a typo?

collection_requirements.rst
```diff
@@ -337,10 +337,8 @@ CI Testing
   * Reviewers can block acceptance of a new collection if they don't agree with the ignores.txt entries.
 
 * You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc) of ``ansible-base``/``ansible-core`` that the collection supports. (Usually the ``HEAD`` of the stable-xxx branches.)
-
 * All CI tests MUST run against every pull request and SHOULD pass before merge.
-* All CI tests MUST pass for the commit that releases the collection.
- 
+* At least sanity tests MUST run against a commit that releases the collection (if the collection has integration/unit tests, they SHOULD run too); if they don't pass, the collection will NOT be released.
```
resmo, line 341: +1

would it make sense to even specify which sanity tests to run? e.g. sanity tests of latest 2 stable branches?

collection_requirements.rst
Andersson007: as i know this format is used to link external pages but the tests fail without this

resmo: okay then ;)

felixfontein: I think it should be either `:ref:` with double backticks and no underscore, or no `:ref:`, and then double backticks with single or double underscore.

felixfontein:
How about not dropping the requirement which tests must be run for a release commit, but instead dropping the requirement that all of them must pass, and specify exactly which are allowed to pass under which circumstances?

Andersson007:
> How about not dropping the requirement which tests must be run for a release commit, but instead dropping the requirement that all of them must pass, and specify exactly which are allowed to pass under which circumstances?

Feel free to suggest the wording, i'm not sure i understand. I'd better drop the requirement because of the reasons explained in https://github.com/ansible-collections/overview/pull/206#issuecomment-1149594097

collection_requirements.rst
felixfontein, line 341:
```suggestion
* At least sanity tests MUST run against a commit that releases the collection; if they do not pass, the collection will NOT be released.. If the collection has integration/unit tests, they SHOULD run too; if they do not pass, the errors SHOULD be analyzed to decide whether they should block the release or not.
```

felixfontein:
> How about not dropping the requirement which tests must be run for a release commit, but instead dropping the requirement that all of them must pass, and specify exactly which are allowed to pass under which circumstances?

After discussing this at Wednesday's meeting, I now agree that requiring sanity checks is enough. In case unit and/or integration tests run, I still recommend that in case they do not pass, they should be looked at to investigate whether they block the release or not. This leaves open the door for failures to be ignored (even automatically, because it says 'should'), but right now this has not been resulting in a decreased quality of collection releases. (Skipping sanity tests did though, but that wasn't allowed so far either, and won't be allowed in the future.)

collection_requirements.rst
Andersson007: LGTM, committed, thanks!

collection_requirements.rst
Andersson007: everything should be fine with the original version according to https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html

If i remove the underscore, it returns:
```
Run rstcheck --report-level error *.rst
collection_requirements.rst:52: (ERROR/3) Unknown interpreted text role "ref".
Error! Issues detected.
Error: Process completed with exit code 1.
```

felixfontein: `:ref:` is Sphinx specific, so it won't work in standalone RST files that aren't processed by Sphinx. The latest version in this PR definitely works though, since it is generic RST.

collection_checklist.md
```diff
@@ -43,5 +43,5 @@ Every comment should say whether the reviewer expects it to be addressed, or whe
 - [ ] if `test/sanity/ignore*.txt` exists, it does not contain error codes listed [here](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#ci-testing)
 - [ ] has CI tests up and running against each of the "major versions" of `ansible-base`/`ansible-core` that the collection supports
 - [ ] all CI tests run against every pull request
-- [ ] all CI tests run against a commit that releases the collection; if they don't pass, the collection won't be released
 - [ ] all CI tests run regularly (nightly, or at least once per week)
+- [ ] Sanity tests run against a commit that releases the collection; if they don't pass, the collection won't be released
```
gundalow, line 47: Please add `MUST` (here and points above)

gundalow: Unit/Integration tests SHOULD run and pass. Though aren't required.

collection_requirements.rst
```diff
@@ -337,10 +337,8 @@ CI Testing
   * Reviewers can block acceptance of a new collection if they don't agree with the ignores.txt entries.
 
 * You MUST run CI against each of the "major versions" (2.10, 2.11, 2.12, etc) of ``ansible-base``/``ansible-core`` that the collection supports. (Usually the ``HEAD`` of the stable-xxx branches.)
-
 * All CI tests MUST run against every pull request and SHOULD pass before merge.
-* All CI tests MUST pass for the commit that releases the collection.
- 
+* At least sanity tests MUST run against a commit that releases the collection; if they do not pass, the collection will NOT be released. If the collection has integration/unit tests, they SHOULD run too; if they do not pass, the errors SHOULD be analyzed to decide whether they should block the release or not.
```
gundalow, line 341: Split into two bullet points 
* sanity
* unit/integration

collection_checklist.md
Andersson007: Added `MUST`. Didn't add unit/integration as they aren't required and i think it makes sense to have only necessary things in the checklist

gundalow (approved):

Andersson007:
thanks everyone!

PR #212, opened 2022-07-31
Make it clear that code outside plugins/ can have other free software licenses as well

felixfontein:
##### SUMMARY
Right now, the rules state that *"[a]ll other code must be under the GPL-3.0-or-later"*. From the explanation it is clear that this sentence is about all other code in plugins/, but it is not formulated that way.

The current formulation implies for example that all unit tests, helper scripts, or code files included in roles, playbooks, or integration tests must be licensed under GPLv3+. Especially for unit tests this can be problematic; for example community.docker now vendors some code from Docker SDK for Python (Apache 2.0 licensed) including unit tests for that code. The unit tests are also licensed under Apache 2.0, but this would violate a strict interpretation of the inclusion requirements.

This PR proposes an improved formulation that makes it clear that other code **outside `plugins/`** can also use other licenses.

(Ref: https://github.com/ansible-collections/community.docker/pull/433#discussion_r933993684)

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
inclusion requirements

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -253,11 +253,11 @@ ways:
                <https://opensource.org/licenses/BSD-2-Clause>`_ license to make it possible for
                third-party modules which are licensed incompatibly with the GPLv3 to use them.
                Please consider this use case when licensing your own ``module_utils``.
-:Code outside plugins/: Code outside `plugins/`, in particular unit tests, must be licensed with
-                        a free software license that is compatible with the
-                       `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
-                       Note that such code that imports GPL licensed code must be GPL licensed
-                       itself as well.
+:Code outside ``plugins/``: Code outside ``plugins/``, in particular unit tests, must be licensed with
+                            a free software license that is compatible with the
+                            `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
+                            Note that such code that imports GPL licensed code must be GPL licensed
+                            itself as well.
 :All other code: All other code must be under the `GPL-3.0-or-later
                  <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
                  inside of the Ansible controller process which is licensed under the GPLv3+ and
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -263,7 +263,9 @@ ways:
                  `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_,
                  provided that such code does not import any other code that is licensed under
                  the ``GPL-3.0-or-later``. If the file does import other ``GPL-3.0-or-later`` code,
-                 then it must similarly be licensed under ``GPL-3.0-or-later``.
+                 then it must similarly be licensed under ``GPL-3.0-or-later``. Note that unit tests
+                 often import code from ansible-core, plugins, module utils, or modules, and such code
+                 is often licensed under ``GPL-3.0-or-later``.
 :Non code content: At the moment, these must also be under the `GPL-3.0-or-later       
                    <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
 
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -256,8 +256,8 @@ ways:
 :All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
                                  <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
                                  are run inside of the Ansible controller process which is licensed under
-                                 the ``GPL-3.0-or-later`` and often must import code from the controller.  For these
-                                 reasons, the GPLv3+ must be used.
+                                 the ``GPL-3.0-or-later`` and often must import code from the controller.
+                                 For these reasons, the GPLv3+ must be used.
 :All other code: Code outside ``plugins/``, in particular unit tests, may be licensed under
                  another free software license that is compatible with the
                  `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_,
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -253,17 +253,17 @@ ways:
                <https://opensource.org/licenses/BSD-2-Clause>`_ license to make it possible for
                third-party modules which are licensed incompatibly with the GPLv3 to use them.
                Please consider this use case when licensing your own ``module_utils``.
-:Code outside ``plugins/``: Code outside ``plugins/``, in particular unit tests, must be licensed with
-                            a free software license that is compatible with the
-                            `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
-                            Note that such code that imports GPL licensed code must be GPL licensed
-                            itself as well.
-:All other code: All other code must be under the `GPL-3.0-or-later
-                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
-                 inside of the Ansible controller process which is licensed under the GPLv3+ and
-                 often must import code from the controller.  For these reasons, the GPLv3+ must be
-                 used.
-:Non code content: At the moment, these must also be under the `GPL-3.0-or-later
+:All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
+                                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
+                                 are run inside of the Ansible controller process which is licensed under
+                                 the GPLv3+ and often must import code from the controller.  For these
+                                 reasons, the GPLv3+ must be used.
+:All other code: Code outside ``plugins/``, in particular unit tests, must be licensed with
+                 a free software license that is compatible with the
+                 `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
+                 Note that such code that imports GPL licensed code must be GPL licensed
+                 itself as well.
+:Non code content: At the moment, these must also be under the `GPL-3.0-or-later       
                    <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
 
 Use `this table of licenses from the Fedora Project
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -253,6 +253,11 @@ ways:
                <https://opensource.org/licenses/BSD-2-Clause>`_ license to make it possible for
                third-party modules which are licensed incompatibly with the GPLv3 to use them.
                Please consider this use case when licensing your own ``module_utils``.
+:Code outside plugins/: Code outside `plugins/`, in particular unit tests, must be licensed with
+                        a free software license that is compatible with the
+                       `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
+                       Note that such code that imports GPL licensed code must be GPL licensed
+                       itself as well.
 :All other code: All other code must be under the `GPL-3.0-or-later
                  <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
                  inside of the Ansible controller process which is licensed under the GPLv3+ and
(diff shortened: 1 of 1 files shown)


collection_requirements.rst
```diff
@@ -262,6 +262,11 @@ ways:
                <https://opensource.org/licenses/BSD-2-Clause>`_ license to make it possible for
                third-party modules which are licensed incompatibly with the GPLv3 to use them.
                Please consider this use case when licensing your own ``module_utils``.
+:Code outside plugins/: Code outside `plugins/`, in particular unit tests, must be licensed with
+                        a free software license that is compatible with the
+                       `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.
+                       Note that such code that imports GPL licensed code must be GPL licensed
+                       itself as well.
```
gotmax23, line 269: This does not seem to be formatted correctly. The indentation is really wonky in the rendered version.

gotmax23:
I agree with the premise of this change. Thank you for proposing it, @felixfontein! I think `All other code` should be clarified to something like `All other plugins` and then the order of rows 3 and 4 should be switched.

collection_requirements.rst
felixfontein: Fixed in 964631699a266c6ccd3e4348b2d164a2a3a91e45.

felixfontein:
> I agree with the premise of this change. Thank you for proposing it, @felixfontein! I think `All other code` should be clarified to something like `All other plugins` and then the order of rows 3 and 4 should be switched.

I implemented that in 3c36d09a90e92b9760b2c841e6d4e344148fe7a3.

collection_requirements.rst
```diff
@@ -262,11 +262,16 @@ ways:
...
-                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
-                 inside of the Ansible controller process which is licensed under the GPLv3+ and
-                 often must import code from the controller.  For these reasons, the GPLv3+ must be
-                 used.
+:All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
+                                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
+                                 are run inside of the Ansible controller process which is licensed under
+                                 the GPLv3+ and often must import code from the controller.  For these
+                                 reasons, the GPLv3+ must be used.
+:All other code: Code outside ``plugins/``, in particular unit tests, must be licensed with
```
gotmax23, line 270: Sorry for not batching up my feedback so you could apply it all at once.

```suggestion
:All other code: Code outside ``plugins/``, in particular unit tests for ``modules`` and ``module_utils``, must be licensed with
```

felixfontein: No problem about batching :)

I'm not sure why unit tests for modules and module_utils should be singled out here. Other unit tests also have to be licensed with a free software license (though very likely with GPLv3+, at least the way I interpret it, since IANAL).

gotmax23: Well, wouldn't the tests for the other plugins import the code from the other plugins that have to be GPL-licensed?

felixfontein: Yes, but (there's always a but ;) ) you can also have Python files there that do not import non-module/module_utils plugins. For example you could have a Apache 2.0 licensed file that provides some utility that is only used by a test for an inventory plugin, and thus located right next to it. Or you can have generic unit test helpers in `tests/unit/` outside of `tests/unit/plugins/modules/` and `tests/unit/plugins/module_utils/` (like the mock import helper) that are licensed under another license than GPLv3+.

gotmax23: Maybe it makes sense to phrase it like so?

```
:All other code: Code outside ``plugins/``, in particular unit tests [for ``modules`` and ``module_utils``], may be licensed under
                 another free software license that is compatible with the
                 `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_,
                 provided that such code that does not import any code that is licensed under
                 the ``GPL-3.0-or-later``.
```

(The bracketed part would either be removed or included without the brackets)

I don't feel strongly either way, but I think my suggestion might be clearer.

gotmax23:
What is the process for getting this approved? Can the SC just vote on it or does it have to go through legal?

felixfontein:
> What is the process for getting this approved? Can the SC just vote on it or does it have to go through legal?

I think a SC vote suffices, since this doesn't really change much. (Assuming nobody from SC disagrees and asks for a review by RH legal... I really don't think that will happen though.)

collection_requirements.rst
felixfontein: Hmm, I think this is only easier to understand for folks who know that code importing GPLv3+ licensed code needs to be GPLv3+ licensed as well. How about adding something like "Code satisfying the latter must be licensed under the `GPL-3.0-or-later`."?

gotmax23: I've edited the suggestion.

felixfontein: Applied as 0a1b0f352485e1ddbbac944130499b2b2a205440 (without the square brackets).

collection_requirements.rst
```diff
@@ -262,11 +262,17 @@ ways:
...
                Please consider this use case when licensing your own ``module_utils``.
-:All other code: All other code must be under the `GPL-3.0-or-later
-                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
-                 inside of the Ansible controller process which is licensed under the GPLv3+ and
-                 often must import code from the controller.  For these reasons, the GPLv3+ must be
-                 used.
+:All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
+                                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
+                                 are run inside of the Ansible controller process which is licensed under
+                                 the GPLv3+ and often must import code from the controller.  For these
```
gotmax23, line 268:
```suggestion
                                 the ``GPL-3.0-or-later`` and often must import code from the controller.  For these
```

gotmax23:
Besides that one other change, this looks good to me.

briantist (approved):

collection_requirements.rst
```diff
@@ -262,11 +262,17 @@ ways:
...
-                 used.
+:All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
+                                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
+                                 are run inside of the Ansible controller process which is licensed under
+                                 the ``GPL-3.0-or-later`` and often must import code from the controller.  For these
+                                 reasons, the GPLv3+ must be used.
+:All other code: Code outside ``plugins/``, in particular unit tests, may be licensed under
+                 another free software license that is compatible with the
+                 `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_,
+                 provided that such code that does not import any other code that is licensed under
```
samccann, line 273:
```suggestion
                 provided that such code  does not import any other code that is licensed under
```

Andersson007 (approved):

collection_requirements.rst
mattclay: Since unit tests import modules, plugins or ansible-core code, it's probably worth clarifying that here.

mattclay:
Since unit tests import modules, plugins or ansible-core code, it's probably worth clarifying that.

[repeating since my original statement was attached to a now hidden comment]

samccann (approved):
LGTM

collection_requirements.rst
```diff
@@ -262,11 +262,19 @@ ways:
...
-                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
-                 inside of the Ansible controller process which is licensed under the GPLv3+ and
-                 often must import code from the controller.  For these reasons, the GPLv3+ must be
-                 used.
+:All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
+                                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
+                                 are run inside of the Ansible controller process which is licensed under
+                                 the ``GPL-3.0-or-later`` and often must import code from the controller.
+                                 For these reasons, the GPLv3+ must be used.
+:All other code: Code outside ``plugins/``, in particular unit tests, may be licensed under
```
mattclay, line 270: This section is long enough that I'm concerned someone will see the "in particular unit tests" part and not notice the caveat at the end regarding unit tests. Perhaps it would be adequate to remove the "in particular unit tests" part of the sentence?

mattclay: Or perhaps unit tests should have their own section ahead of "all other code" so we can more clearly address the topic there without making this section longer.

felixfontein: How about rewriting this section as
```
:All other code: Code outside ``plugins/`` may be licensed under another free software license that is compatible
                 with the `GPL-3.0-or-later <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_,
                 provided that such code does not import any other code that is licensed under
                 the ``GPL-3.0-or-later``. If the file does import other ``GPL-3.0-or-later`` code,
                 then it must similarly be licensed under ``GPL-3.0-or-later``. Note that this applies in
                 particular to unit tests; these often import code from ansible-core, plugins, module utils,
                 or modules, and such code is often licensed under ``GPL-3.0-or-later``.
```

mattclay: I think that would be adequate.

felixfontein: I've added it in 626b67e21400d9670308e99b4535277662025eb7. Thanks!

felixfontein:
Any more comments? Or should we start voting on the current state?

felixfontein:
There is now an active vote on this topic: https://github.com/ansible-community/community-topics/discussions/142

collection_requirements.rst
```diff
@@ -262,11 +262,18 @@ ways:
...
-:All other code: All other code must be under the `GPL-3.0-or-later
-                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins are run
-                 inside of the Ansible controller process which is licensed under the GPLv3+ and
-                 often must import code from the controller.  For these reasons, the GPLv3+ must be
-                 used.
+:All other code in ``plugins/``: All other code in ``plugins/`` must be under the `GPL-3.0-or-later
+                                 <https://www.gnu.org/licenses/gpl-3.0-standalone.html>`_.  These plugins
+                                 are run inside of the Ansible controller process which is licensed under
+                                 the ``GPL-3.0-or-later`` and often must import code from the controller.
+                                 For these reasons, the GPLv3+ must be used.
```
gotmax23, line 269: `s/GPLv3+/GPL-3.0-or-later/`

felixfontein:
(I rebased after #216 was merged, no content changes.)

collection_requirements.rst
felixfontein: Implemented in 48da08635ce9bdf7a571572610f17f3dbaae446b.

Andersson007 (commented):
@felixfontein thanks for the clarifications!
I have 2 questions:
1. Do i understand correctly that if a collection is fully licensed with a GPL compatible license and contains, say, `plugins/actions/..` or anything else besides `modules` and `module_utils` in the `plugins` directory, the maintainers must re-license it with GPLv3+, otherwise we can't include/have such a collection included?
2. Should we use uppercase modal verbs as we do in other paragraphs?

Thanks

felixfontein:
>     1. Do i understand correctly that if a collection is fully licensed with a GPL compatible license and contains, say, `plugins/actions/..` or anything else besides `modules` and `module_utils` in the `plugins` directory, the maintainers must re-license it with GPLv3+, otherwise we can't include/have such a collection included?

Yes. But that's not related to including the collection into Ansible, it's a result of the GPL already. Action plugins (and other plugin types, except a few of them) import code from ansible-core, and thus these plugins **must** be licensed under GPLv3 as well.

felixfontein:
> Should we use uppercase modal verbs as we do in other paragraphs?

Hmmm, I'm not sure whether this is a good idea to change since the vote already started.

Andersson007:
> > Should we use uppercase modal verbs as we do in other paragraphs?
> 
> Hmmm, I'm not sure whether this is a good idea to change since the vote already started.

It doesn't change the meaning, just makes it a bit more recognizable (anyway it's not an obstacle to merge it as-is and we can do it later in another PR). Let's continue without changing it here

felixfontein:
This PR has been approved, see https://github.com/ansible-community/community-topics/issues/123#issuecomment-1275020592.

felixfontein:
Thanks everyone! :)

PR #214, opened 2022-08-14
Remove officially unmaintained collections

mariolenz:
##### SUMMARY
The current process to identify and remove an unmaintained collection looks very much focused on _abandoned_ collections. But I think we should have a way to remove a collection that announces itself as being unmaintained quicker.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
removal_from_ansible.rst

##### ADDITIONAL INFORMATION
ansible-community/community-topics#124

cc @felixfontein


felixfontein:
I marked this as WIP since this needs to be voted on before it can be merged.

mariolenz:
> I marked this as WIP since this needs to be voted on before it can be merged.

I've added ansible-community/community-topics#130. Maybe we can discuss this tomorrow and, if people agree, I'll start a vote.

removal_from_ansible.rst
```diff
@@ -125,8 +125,16 @@ Collections that have been announced deprecated / unmaintained by their maintain
...
 ~~~~~~~
 
-#. `Announce upcoming removal from the next major Ansible release <announce_removal_>`_.
-#. `Remove collection from the next major Ansible release <remove_collection_>`_.
+If the latest stable Ansible version is X.something:
+
+#. Announce upcoming removal in the Ansible changelog (``https://github.com/ansible-community/ansible-build-data/blob/main/<X>/changelog.yaml``).
+#. Remove from ``ansible.in`` (``https://github.com/ansible-community/ansible-build-data/blob/main/<X+1>/ansible.in``).
+#. Remove from ``collection-meta.yaml`` (``https://github.com/ansible-community/ansible-build-data/blob/main/<X+1>/collection-meta.yaml``).
+#. Document actual removal for the next Ansible release in the changelog (``https://github.com/ansible-community/ansible-build-data/blob/main/<X+1>/changelog.yaml``).
```
felixfontein, line 133: The goal of referring to the general processes inside this document was to avoid repeating these instructions everywhere.

mariolenz: I understand, but referring to `announce removal` and `remove collection` looks like there are two separate steps. I think we can do it in just one step. The process I propose is not simply `announce removal` + `remove collection` ;-)

Additionally, `Announcing upcoming removal` says to `announce upcoming removal in the collection's issue tracker`. Which doesn't make sense if the collection is officially unmaintained. And, actually, wouldn't even be possible if the collection has been archived.

flowerysong: I don't see the point of two divergent processes for doing the exact same thing, or duplicating documentation details in order to do so. If you think that in some circumstances announcing it on the collection's issue tracker won't be possible, just add an "if possible" on the end of the existing documentation for that step.

mariolenz: @felixfontein @flowerysong What do you think about my latest commit? Does it look OK to you now?

removal_from_ansible.rst
```diff
@@ -119,8 +119,19 @@ Process
...
+Collections that have been announced deprecated / unmaintained by their maintainers
+-----------------------------------------------------------------------------------
+
+Process
+~~~~~~~
+
+#. `Announce upcoming removal from the next major Ansible release <announce_removal_>`_.
+#. `Remove collection from the next major Ansible release <remove_collection_>`_.
+
+See `<https://github.com/ansible-community/ansible-build-data/pull/374/files>`__ for an example on how to this in just one PR in ``ansible-build-data``.
```
Andersson007, line 131:
```suggestion
See `the example pull request <https://github.com/ansible-community/ansible-build-data/pull/374/files>`_ in the ``ansible-build-data`` repository to learn how to remove the collection.
```

mariolenz:
@felixfontein @flowerysong I'm referring to the general processes again. What do you think about this PR now?

felixfontein (approved):
Thanks, looks good to me!

briantist (approved):

russoz (approved):
LGTM

Andersson007 (approved):

removal_from_ansible.rst
```diff
@@ -119,8 +119,19 @@ Process
...
-Identifying and removing an unmaintained collection
----------------------------------------------------
+Collections that have been announced deprecated / unmaintained by their maintainers
+-----------------------------------------------------------------------------------
+
+Process
+~~~~~~~
+
+#. `Announce upcoming removal from the next major Ansible release <announce_removal_>`_.
+#. `Remove collection from the next major Ansible release <remove_collection_>`_.
```
samccann, line 129: Picking a corner case here - What if a collection owner decides it's deprecated/unmaintained late in an Ansible development cycle?  Say foo.bar collection is deprecated a week before the Ansible 10 alpha - do we still remove it in Ansible 10?

mariolenz: @samccann Good point, we should clarify this.

In this case, I think we should remove the collection since it's only an alpha release. That's not late, it's early in the development cycle. But probably not once there's a pre-release that comes with a feature freeze (Ansible-10.0.0 beta1).

What do you think about:

```
If the current major release is X and there hasn't been a feature freeze release of the next major version X+1:

#. `Announce upcoming removal from the (X+1).0.0 Ansible release <announce_removal_>`_.
#. `Remove collection from the (X+1).0.0 Ansible release <remove_collection_>`_.

If there already has been a feature freeze release of the next major version X+1:

#. `Announce upcoming removal from the (X+2).0.0 Ansible release <announce_removal_>`_.
#. `Remove collection from the (X+2).0.0 Ansible release <remove_collection_>`_.
```

felixfontein: Or simply define a new variable `Y` which is `X+1` or `X+2`, and then use `Y`, to avoid having the list of instructions twice :)

mariolenz: @samccann Or, as @felixfontein suggested, simplify this using a variable `Y`:

```
If the current major release is X and there hasn't been a feature freeze release of the next major version X+1, remove the collection from Y=(X+1).0.0.
If there already has been a feature freeze release of the next major version X+1, remove the collection from Y=(X+2).0.0.

#. `Announce upcoming removal from the (Y Ansible release <announce_removal_>`_.
#. `Remove collection from the Y Ansible release <remove_collection_>`_.
```

felixfontein: There's now an opening bracket in the first `#.` line that shouldn't be there, but besides from that, LGTM.

mariolenz:
@samccann What do you think about my last commit / this PR now?

samccann (approved):

samccann:
LGTM

removal_from_ansible.rst
```diff
@@ -119,8 +119,22 @@ Process
 Unmaintained collections
 ========================
 
-Identifying and removing an unmaintained collection
----------------------------------------------------
+Collections that have been announced deprecated / unmaintained by their maintainers
```
flowerysong, line 122:
```suggestion
Removing a collection that has been explicitly deprecated or abandoned by its former maintainers
```

removal_from_ansible.rst
```diff
@@ -119,8 +119,22 @@ Process
...
+
+If the current major release is X and there hasn't been a feature freeze release of the next major version X+1, remove the collection from Y=(X+1).0.0.
+If there already has been a feature freeze release of the next major version X+1, remove the collection from Y=(X+2).0.0.
+
+#. `Announce upcoming removal from the Y Ansible release <announce_removal_>`_.
+#. `Remove collection from the Y Ansible release <remove_collection_>`_.
+
+See `the example pull request <https://github.com/ansible-community/ansible-build-data/pull/374/files>`_ in the ``ansible-build-data`` repository to learn how to remove the collection.
+
+Identifying and removing an unmaintained collection that has not been deprecated by their maintainers
```
flowerysong, line 136:
```suggestion
Identifying and removing an unmaintained collection that has not been deprecated by its maintainers
```

flowerysong (commented):
Really just a number agreement/grammar nit, though I also tried to make the first heading's structure more like other headings at this level.

removal_from_ansible.rst
mariolenz: I don't like "former maintainers" here. When they deprecate or abandon the collection, they are still the maintainers. They are only _former maintainers_ afterwards.

What I mean is: Only _current_ maintainers can deprecate or abandon a collection. What do you think about my [last commit](https://github.com/ansible-collections/overview/pull/214/commits/81c1150dc3970d46f23c503e0e9822db804fd36a)?

flowerysong: It's not a big deal either way, but when this process is taking place they are the former maintainers so it feels weird to call them the maintainers. The collection currently has no maintainers, or else it would not be eligible for this process.

mariolenz: @Andersson007 @briantist @felixfontein @russoz @samccann What do you think, "former maintainers" or just "maintainers"? Personally, I don't like the "former" but I'm open to both.

felixfontein: Deprecated collections can still be maintained. Only when they are unmaintained or abandoned (is there actually a difference between the two?) they no longer have maintainers. So depending on context, both can be the right wording. So how about writing `(former) maintainers`?

mariolenz:
The [vote on merging this](https://forum.ansible.com/t/4204/3) has been accepted.

PR #215, opened 2022-09-29
Inclusion: Update wording of temporary files to exclude editor files

briantist:
##### SUMMARY

Resolves: https://github.com/ansible-community/community-topics/issues/138

In a bit of irony, my own editor config has "fixed" some whitespace issues here 😅 so, I've separated the content changes into a separate commit for easier review: 
- aa33de2b21f14ac714966a14962a1bb4eba20be1

(I can also rebase to remove the remove the whitespace changes after the content is decided on, if that's preferable)

##### ISSUE TYPE

- Docs Pull Request


felixfontein:
I think it's best to move the whitespace changes to another PR, because that one we can just merge, while this one needs a vote ;-)

collection_checklist.md
```diff
@@ -37,7 +37,7 @@ Note for reviewers: If you don't know how to check any of the points below, plea
 - [ ] `meta/runtime.yml` defines the minimal version of Ansible which the collection works with
 - [ ] has changelog, preferably with `changelogs/changelog.yaml`
 - [ ] collection repository should not contain any large objects (binaries) comparatively to the current Galaxy tarball size limit of 20 MB. For example, package installers for testing purposes shouldn't be added.
-- [ ] collection repository should not contain any unnecessary files like, for example, temporary files created by development tools. Temporary files MUST be added to `.gitignore`.
+- [ ] collection repository should not contain any unnecessary files like, for example, temporary files. Temporary files should be added to `.gitignore`.
```
gotmax23, line 40: Why not must?

briantist: There's no such corresponding must in the full set of rules, and to @felixfontein 's comment in https://github.com/ansible-community/community-topics/issues/138 , it's hard to use a MUST if we're being somewhat ambiguous about what counts as temporary files.

I was even debating taking out the `.gitignore` comment, or rewording to something like "consider adding to `.gitignore`" or whatever.

Happy to reconsider; if we want this as must, perhaps we should add it to the full criteria as well?

felixfontein: I'm not sure how "must" actually makes sense here. Would that mean that every possible kind of temporary file must be mentioned in `.gitignore`? I'm wondering how one would actually do that :) In that case "should" is better than "must", but "consider adding to `.gitignore`" is even better than "should".

briantist:
> I think it's best to move the whitespace changes to another PR, because that one we can just merge, while this one needs a vote ;-)

I could do that if it will be merged in short order and I can rebase this; I'm also fine leaving it in this one (and waiting for the vote).
If I remove it from here (without it being merged/rebases) I will have to deal with it every time I save a new change.

collection_checklist.md
gotmax23: > it's hard to use a MUST if we're being somewhat ambiguous about what counts as temporary files.

That's a good point. I think `should` makes sense here then.

felixfontein:
> I could do that if it will be merged in short order and I can rebase this

That should happen :)

briantist:
@felixfontein 
- https://github.com/ansible-collections/overview/pull/216

briantist:
thanks @felixfontein I have rebased and there are no more whitespace-only changes in this PR

briantist:
@Andersson007 would you like to suggest any changes here?

Andersson007 (approved):
@briantist LGTM thanks!

Andersson007:
I'm not sure how to proceed, folks thoughts?
option 1: merge it as it just removes confusion
option 2: vote in the topics

briantist:
vote is open!
- https://github.com/ansible-community/community-topics/discussions/146

Andersson007:
Thanks @briantist and everyone involved!

PR #217, opened 2022-10-13
removal_from_ansible.rst: add paragraph about removal collections violating the requirements

Andersson007:
##### SUMMARY
removal_from_ansible.rst: add paragraph about removal collections violating the requirements

to explicitly cover cases that don't belong to unmaintained/broken collections

related discussion https://github.com/ansible-community/community-topics/issues/150

### DIFF
--- a/removal_from_ansible.rst
+++ b/removal_from_ansible.rst
@@ -188,10 +188,9 @@ Process
 #. Any community member who finds a collection that violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ may file an issue against said collection's repository. If the reporter is unsure whether something constitutes a violation or believes that the apparently violated guideline is unclear, they should consult with the steering committee by filing a `community topic <https://github.com/ansible-community/community-topics/issues>`_ before proceeding.
 #. The issue filed against the collection's repository should include the following information:
 
-  * References to the corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates
-  * Actions collection maintainers need to do to make the collection comply with the requirements
-  * A link to the Collection Requirements violations tracking issue (https://github.com/ansible-community/community-topics/issues/PLACEHOLDER)
-
+  * References to the corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates.
+  * Actions collection maintainers need to do to make the collection comply with the requirements.
+  * A link to the Collection Requirements violations tracking `issue <https://github.com/ansible-community/community-topics/issues/>`_.
 #. A default term for the collection to solve the issue is four weeks since the issue was created. It can vary depending on a requirement violated, SC opinions or other circumstances.
 #. If the violation is not fixed or there is a disagreement between the reporter and the maintainers, the reporter or another person creates a `community topic <https://github.com/ansible-community/community-topics/issues>`_.
 #. Two SC members check the reported circumstances and confirm in the topic that the violation is present.
--- a/removal_from_ansible.rst
+++ b/removal_from_ansible.rst
@@ -169,7 +169,7 @@ There is no simplified process. Once the collection has been removed from Ansibl
 Collections not satisfying the Collection Requirements
 ======================================================
 
-A collection can be removed from the package if it violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ without making enough effort to mitigate these violations.
+A collection can be removed from the package if it violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ without resolving the violations within the time allowed.
 
 This section is not applicable to cases of broken or unmaintained collections. Instead, see the corresponding paragraphs of this document.
 
--- a/removal_from_ansible.rst
+++ b/removal_from_ansible.rst
@@ -186,7 +186,8 @@ Process
 ~~~~~~~
 
 #. A community person who found a violation creates a `community topic <https://github.com/ansible-community/community-topics/issues>`_.
-#. If SC members generally agree that the violation is present, a community person creates an issue in the collection's repository. The issue contains:
+#. Two SC members check the reported circumstances and confirm in the topic that the violation is present.
+#. A community person creates an issue in the collection's repository. The issue contains:
 
   * References to corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates.
   * Actions collection maintainers need to do to make the collection satisfying the requirements.
--- a/removal_from_ansible.rst
+++ b/removal_from_ansible.rst
@@ -18,9 +18,9 @@ General processes
 
 The general process of removing a collection follows these steps:
 
-#. announcing an upcoming removal of a collection
-#. removing the collection
-#. when appropriate, re-adding the collection
+#. Announcing an upcoming removal of a collection.
+#. Removing the collection.
+#. When appropriate, re-adding the collection.
 
 
 .. _announce_removal:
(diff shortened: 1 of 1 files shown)


removal_from_ansible.rst
```diff
@@ -165,3 +165,57 @@ Re-adding collection to Ansible
 -------------------------------
 
 There is no simplified process. Once the collection has been removed from Ansible Y.0.0, it needs to go through the full inclusion process to be re-added to the Ansible package. Exceptions are only possible if the Steering Committee votes on them. The Steering Committee can approve or deny a fast re-entry without going through the full review process.
+
+Collections not satisfying the Collection Requirements
+======================================================
+
+A collection can be removed from the package if it violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_.
```
felixfontein, line 172: How about something like:
```suggestion
A collection can be removed from the package if it violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ without making enough effort to mitigate these violations.
```
Besides that, looks good!

Andersson007: lgtm, committed, thanks!

mariolenz (approved):
LGTM

removal_from_ansible.rst
```diff
@@ -165,3 +165,57 @@ Re-adding collection to Ansible
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+#. A collection violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_.
+#. Collection maintainers have not fixed the violations and have not released a fixed version of the collection within the time period established by this document.
+
+Process
+~~~~~~~
+
+#. A community person who found a violation creates a `community topic <https://github.com/ansible-community/community-topics/issues>`_.
+#. If SC members generally agree that the violation is present, a community person creates an issue in the collection's repository. The issue contains:
```
gotmax23, line 189: Why does the SC need to adjudicate whether something constitutes a violation before a community member can create an issue? While the steering committee should be notified of violations, I think only the actual removal process should require a vote.

felixfontein: I don't think this is meant as a vote, more that at least two SC members agree that this really is a violation (and not something which should be valid, but the formation of the rules needs to be improved, or a misunderstanding, or ...). Basically a small sanity check whether more than one person who has some knowledge with the rules thinks that they are violated.

Andersson007: yes, no vote was implied, just you know something like when you report say concerns in the meeting and couple of SC say "yep, it seems to be a violation", that's enough to start the process (to save issue reporter's and others' time in case it's a false warning). I'll try to come up with a better wording. thanks!

Andersson007: @gotmax23 @felixfontein changed, PTAL

felixfontein: Sounds good to me!

gotmax23: I just meant I don't think the steering committee needs to be involved with every potential violation. What do you all think about making filling an issue more of a recommendation for when things are unclear, so we don't need extra bureaucracy and issues when it's a cut and dry violation? This requirement wasn't there before, and I'm not convinced it should be added as part of this change.

gotmax23: For instance, a few months ago, I found a violation of the licensing guidelines in community.hashi_vault. I just filed https://github.com/ansible-collections/community.hashi_vault/issues/276 and @briantist fixed it. If this policy was in effect then, wouldn't I have had to file a community-topics issue first? Or am I still misunderstanding?

felixfontein: Hmm, I was more thinking about cases where not all involved (reporter and collection maintaines) are sure that this is a violation. Maybe it should be formulated so that the SC is only involved if there is disagreement between the reporter and the maintainers that this is a violation, and/or that it needs to be fixed?

Andersson007: @gotmax23 @felixfontein good points! Re-formulated, PTAL

removal_from_ansible.rst
```diff
@@ -165,3 +165,58 @@ Re-adding collection to Ansible
...
+Process
+~~~~~~~
+
+#. A community person who found a violation creates an issue in the collection's repository. The issue contains:
+
+  * References to corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates.
+  * Actions collection maintainers need to do to make the collection satisfying the requirements.
+#. A default term for the collection to solve the issue is four weeks since the issue was created. It can vary depending on a requirement violated, SC opinions or other circumstances.
+#. If the violation is not fixed or there is a disagreement between the reporter and the maintainers, the reporter or another person creates a `community topic <https://github.com/ansible-community/community-topics/issues>`_.
+#. Two SC members check the reported circumstances and confirm in the topic that the violation is present.
```
felixfontein, line 194: I'm wondering whether we still need this, or in which form precisely. The basic idea is to have a 'sanity check' whether this is actually something that needs to be fixed in the collection, or whether it is a) a wrong interpretation of the rule, or b) rather the rule should be changed than the violation, or c) something else. This step should be deciding between "let's discuss this further" and "let's start a vote (with a binary result)". What do you all think, should this be updated/modified (if yes: how?) to explain that better?

Andersson007: cc @mariolenz @gotmax23 ^

gotmax23: Yeah, I kind of agree with @felixfontein. This is how we conduct all votes; there's a few days for folks to leave comments and discuss the merits, and then we start the vote if there's no significant opposition. I don't have strong feelings about whether this step should be changed or removed, though, so I'll defer to others.

removal_from_ansible.rst
```diff
@@ -165,3 +165,58 @@ Re-adding collection to Ansible
...
+#. A collection violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_.
+#. Collection maintainers have not fixed the violations and have not released a fixed version of the collection within the time period established by this document.
+
+Process
+~~~~~~~
+
+#. A community person who found a violation creates an issue in the collection's repository. The issue contains:
+
+  * References to corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates.
+  * Actions collection maintainers need to do to make the collection satisfying the requirements.
```
gotmax23, line 191: This recommends that community members create an community topic if they're unsure whether something constitutes a violation. I think it would also make sense to create a tracking issue for collection violations that people can mention when filling issues so that Github links them. Doing so will allow us to track the situation and identify patterns without making the process overly burdensome. I'm not super attached to the tracking issue idea, but I think it's a good middle ground.


```suggestion
#. Any community member who finds a collection that violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ may file an issue against said collection's repository. If the reporter is unsure whether something constitutes a violation or believes that the apparently violated guideline is unclear, they should consult with the steering committee by filing a `community topic <https://github.com/ansible-community/community-topics/issues>`_ before proceeding.
#. The issue filed against the collection's repository should include the following information:

  * References to the corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates
  * Actions collection maintainers need to do to make the collection comply with the requirements
  * A link the the Collection Requirements violations tracking issue (https://github.com/ansible-community/community-topics/issues/PLACEHOLDER)

```

removal_from_ansible.rst
gotmax23: Yeah, I kind of agree with @felixfontein. This is how we conduct all votes; there's a few days for folks to leave comments and discuss the merits, and then we start the vote if there's no significant opposition. I don't have strong feelings about whether this step should be changed or removed, though, so I'll defer to others.

removal_from_ansible.rst
mariolenz: @gotmax23 

> A link the the Collection Requirements violations tracking issue (https://github.com/ansible-community/community-topics/issues/PLACEHOLDER)

I think this should be "to the" instead of "the the", shouldn't it?

gotmax23: Indeed. I've fixed it.

Andersson007: I like the proposed wording, much better now, thanks! Committed.
I added periods and changed the link in the last bullet point. PTAL

gotmax23: The idea was that we would create a single tracking issue so people could add something like `Relates: github.com/ansible-community/community-topics/issue/12345` to the end of the issues filled against collections. Then, every issue that included that link would show up on the tracking issue's page. PLACEHOLDER was meant to be replaced with the ID of that new issue.

In Fedora, we liberally use Bugzilla's support for tracking bugs, but not everyone uses that, so I should've explained myself better :smile:. This is me trying to replicate that with Github's functionality. I wanted find an easy way for the SC to keep track of these violations without making people create issues against two different repositories. I see the issue linking approach might be confusing, won't send notifications, and won't work for collections outside of Github, so maybe we should scrap it...

Andersson007: > The idea was that we would create a single tracking issue so people could add something like Relates: github.com/ansible-community/community-topics/issue/12345 to the end of the issues filled against collections. Then, every issue that included that link would show up on the tracking issue's page. PLACEHOLDER was meant to be replaced with the ID of that new issue.

@gotmax23 thanks for the clarifications. If I understand correctly, it's gonna be like a permanent issue, right? If yes, it works for me (regarding the second paragraph: we could try and change it later if needed).
How about @mariolenz @felixfontein ?
If everyone is happy with the tracking issue, let's change it when the issue is created (i believe we should vote as usual before creating the issue)?

gotmax23: > If I understand correctly, it's gonna be like a permanent issue, right?

Exactly.

> (i believe we should vote as usual before creating the issue)?

I think it can be one vote. If the motion passes, we can then change the link and create the issue.

Andersson007: > I think it can be one vote. If the motion passes, we can then change the link and create the issue.

SGTM

removal_from_ansible.rst
```diff
@@ -165,3 +165,60 @@ Re-adding collection to Ansible
 -------------------------------
 
 There is no simplified process. Once the collection has been removed from Ansible Y.0.0, it needs to go through the full inclusion process to be re-added to the Ansible package. Exceptions are only possible if the Steering Committee votes on them. The Steering Committee can approve or deny a fast re-entry without going through the full review process.
+
+Collections not satisfying the Collection Requirements
+======================================================
+
+A collection can be removed from the package if it violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ without making enough effort to mitigate these violations.
```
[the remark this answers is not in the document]
Andersson007: @jctanner i do like the suggestion, much better, changed, thanks!

removal_from_ansible.rst
```diff
@@ -165,3 +165,60 @@ Re-adding collection to Ansible
...
+
+Process
+~~~~~~~
+
+#. Any community member who finds a collection that violates one or more of the `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ may file an issue against said collection's repository. If the reporter is unsure whether something constitutes a violation or believes that the apparently violated guideline is unclear, they should consult with the steering committee by filing a `community topic <https://github.com/ansible-community/community-topics/issues>`_ before proceeding.
+#. The issue filed against the collection's repository should include the following information:
+
+  * References to the corresponding `Collection Requirements <https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst>`_ the collection violates.
+  * Actions collection maintainers need to do to make the collection comply with the requirements.
+  * A link to the Collection Requirements violations tracking `issue <https://github.com/ansible-community/community-topics/issues/>`_.
```
felixfontein, line 193: Should this only be added if such an issue exists? I'm confused what exactly I should link to when I create such an issue.

removal_from_ansible.rst
```diff
@@ -165,3 +165,60 @@ Re-adding collection to Ansible
...
+#. A default term for the collection to solve the issue is four weeks since the issue was created. It can vary depending on a requirement violated, SC opinions or other circumstances.
+#. If the violation is not fixed or there is a disagreement between the reporter and the maintainers, the reporter or another person creates a `community topic <https://github.com/ansible-community/community-topics/issues>`_.
+#. Two SC members check the reported circumstances and confirm in the topic that the violation is present.
```
felixfontein, line 196: How about:
```suggestion
#. Two SC members check the reported circumstances and confirm in the topic that the violation is present from their point of view, and is one that must be fixed.
```

Andersson007: LGTM

removal_from_ansible.rst
Andersson007: SGTM
@gotmax23 would you like to create the issue and provide the link?

gotmax23: From #ansible-community/2022-10-27.log:
```
* <gotmax[m]> worries that he's overcomplicated the removal from ansible PR
<gotmax[m]> felixfontein andersson007_: Wdyt about just scrapping the tracking issue idea and voting on the rest?

````

I thought about this more, and I don't think it'll work as I envisioned. I'd prefer to leave this out of the policy for now and vote on the rest. Sorry for the noise.

Andersson007: No worries, removed

Andersson007:
Cool, i've just opened a vote https://github.com/ansible-community/community-topics/discussions/158

gotmax23 (approved):

gundalow (approved):
Nicely laid out, thanks for adding this

Andersson007:
thanks everyone involved for your great suggestions!

PR #218, opened 2022-10-13
Clarify SCM and release requirements

mariolenz:
##### SUMMARY

We require that all collections [MUST have a public SCM repository](https://github.com/ansible-collections/overview/blob/24343d3d7723adaf911f2fcc3a69d02bb713fa2a/collection_requirements.rst#repository-management).

Actually, most people will use git nowadays but in theory, they could use any obscure SCM as long as the repo is publicly available.

This means that we might have to deal with a SCM we don't know and our tooling doesn't support. Therefor, I suggest requiring git instead of generic SCM repositories.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
collection_checklist.md
collection_requirements.rst

##### ADDITIONAL INFORMATION
https://github.com/ansible-community/community-topics/issues/148#issuecomment-1277828717

cc @felixfontein

### DIFF
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -280,9 +280,11 @@ licensing and legal concerns that may otherwise affect your code.
 Repository management
 =====================
 
-Every collection MUST have a public git repository and releases of the collection MUST be tagged in this repository.
+Every collection MUST have a public git repository. Releases of the collection MUST be tagged in said repository. This means that releases MUST be ``git tag``ed and that the tag name MUST exactly match the Galaxy version number. Tag names MAY have a ``v`` prefix, but a collection's tag names MUST have a consistent format from release to release.
 
-We are open to add more SCMs once our tooling supports them.
+Additionally, collection artifacts released to Galaxy MUST be built from the sources that are tagged in the collection's git repository as that release. Any changes made during the build process MUST be clearly documented so the collection artifact can be reproduced.
+
+We are open to allowing other SCM software once our tooling supports them.
 
 Branch name and configuration
 -----------------------------
--- a/collection_requirements.rst
+++ b/collection_requirements.rst
@@ -280,7 +280,9 @@ licensing and legal concerns that may otherwise affect your code.
 Repository management
 =====================
 
-Every collection MUST have a public SCM repository and releases of the collection MUST be tagged in this repository.
+Every collection MUST have a public git repository and releases of the collection MUST be tagged in this repository.
+
+We are open to add more SCMs once our tooling supports them.
 
 Branch name and configuration
 -----------------------------
--- a/collection_checklist.md
+++ b/collection_checklist.md
@@ -11,7 +11,7 @@ Note for reviewers: If you don't know how to check any of the points below, plea
 - [ ] has a policy of releasing, versioning and deprecation announced to contributors and users in some way
 - [ ] have a Code of Conduct (CoC) compatible with the [Ansible Code of Conduct (CoC)](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html)
 - [ ] have a publicly available issue tracker that does not require a paid level of service to create an account or view issues
-- [ ] have a public SCM repository, releases of the collection are tagged in this repository
+- [ ] have a public git repository, releases of the collection are tagged in this repository
 
 **Standards and documentation:**
 - [ ] adheres to [semantic versioning](https://semver.org/)


gotmax23:
*Currently*, I don't think we have any tooling that depends on the presence of a Git repository. I'm okay with this change, but I think it should be made in conjunction with the proposed wording and enforcement changes of the release tagging rule (https://github.com/ansible-community/community-topics/issues/148) to clarify the reason for requiring a git repository and so we don't have to vote twice.

mariolenz:
> I'm okay with this change, but I think it should be made in conjunction with the proposed wording and enforcement changes of the release tagging rule ([ansible-community/community-topics#148](https://github.com/ansible-community/community-topics/issues/148)) to clarify the reason for requiring a git repository and so we don't have to vote twice.

I'm not sure what's better, having this as a change on it's own or do it in conjunction with ansible-community/community-topics#148. I'm fine with both options.

collection_requirements.rst
```diff
@@ -280,7 +280,9 @@ licensing and legal concerns that may otherwise affect your code.
 Repository management
 =====================
 
-Every collection MUST have a public SCM repository and releases of the collection MUST be tagged in this repository.
+Every collection MUST have a public git repository and releases of the collection MUST be tagged in this repository.
+
+We are open to add more SCMs once our tooling supports them.
```
gotmax23, line 285: If we're explicitly requiring a git repository, here is how I'd reword my original suggestion. If there's consensus to handle this separately, I can create a separate PR.

```suggestion
Every collection MUST have a public git repository. Releases of the collection MUST be tagged in said repository. This means that releases MUST be ``git tag``ed and that the tag name MUST exactly match the Galaxy version number. Tag names MAY have a ``v`` prefix, but a collection's tag names MUST have a consistent format from release to release.

Additionally, collection artifacts released to Galaxy MUST be built from the sources that are tagged in the collection's git repository as that release. Any changes made during the build process MUST be clearly documented so the collection artifact can be reproduced.

We are open to allowing other SCM software once our tooling supports them.
```

mariolenz:
@gotmax23 

> If there's consensus to handle this separately, I can create a separate PR.

I might be wrong, but I don't think we need to handle this seperately. I've copied your suggestion over to [ansible-community/community-topics#148](https://github.com/ansible-community/community-topics/issues/148#issuecomment-1278519795) so it's not lost and will close this PR.

felixfontein:
I would have just updated this PR with the new suggestion, but we can also have a new one, or first continue discussing it in community-topics. Whatever you prefer :)

mariolenz:
> I would have just updated this PR with the new suggestion, but we can also have a new one, or first continue discussing it in community-topics. Whatever you prefer :-)

@felixfontein @gotmax23 I've re-opened the PR, merged the suggestion and changed the title.

gotmax23 (approved):

felixfontein (approved):
I think this looks good. Maybe let's wait a few more days for feedback, and if there is none start a vote? Or do you think we should already start voting sooner?

mariolenz:
> Maybe let's wait a few more days for feedback, and if there is none start a vote? Or do you think we should already start voting sooner?

I've opened a vote because this has been (basically) discussed in the last community meeting iirc, so people should know about this and already had a couple of days to give feedback.

I have given an end date that's a few days later than the usual 1 week so people have more time to react, though.

collection_requirements.rst
```diff
@@ -280,7 +280,11 @@ licensing and legal concerns that may otherwise affect your code.
 Repository management
 =====================
 
-Every collection MUST have a public SCM repository and releases of the collection MUST be tagged in this repository.
+Every collection MUST have a public git repository. Releases of the collection MUST be tagged in said repository. This means that releases MUST be ``git tag``ed and that the tag name MUST exactly match the Galaxy version number. Tag names MAY have a ``v`` prefix, but a collection's tag names MUST have a consistent format from release to release.
+
+Additionally, collection artifacts released to Galaxy MUST be built from the sources that are tagged in the collection's git repository as that release. Any changes made during the build process MUST be clearly documented so the collection artifact can be reproduced.
+
+We are open to allowing other SCM software once our tooling supports them.
```
briantist, line 287: While I'm not opposed to this sentence, it seems unlikely we'll ever make the tooling support other SCM software unless a collection that uses it exists, and if we're preventing it from being added based on that, I doubt we'd ever support them.

I'd also like to know what tooling specifically.. though it doesn't have to be listed in this doc.

briantist (approved):

collection_requirements.rst
mariolenz: > I'd also like to know what tooling specifically.. though it doesn't have to be listed in this doc.

Probably [collection_prep](https://github.com/ansible-network/collection_prep). I'm not an expert on this tool, but I think it won't work with non-git repositories.

@felixfontein Did you have anything else in mind when saying that [we can potentially add more SCMs to the list once our tooling supports them](https://github.com/ansible-community/community-topics/issues/148#issuecomment-1277837851)?

gotmax23: > I'd also like to know what tooling specifically.. though it doesn't have to be listed in this doc.

See https://github.com/ansible-community/community-topics/issues/148#issuecomment-1276668186. My idea was that we'd set up an automated check to ensure that new collection Galaxy releases are tagged in the respective collection's repository. @felixfontein pointed out that would fall flat if we didn't actually mandate that collections use a git repository.

>  it seems unlikely we'll ever make the tooling support other SCM software unless a collection that uses it exists, and if we're preventing it from being added based on that, I doubt we'd ever support them.

I guess if a new collection wanted to use some other SCM, that's when we'd consider allowing it.

Andersson007:
@mariolenz thanks for the improvement! @gotmax23 @felixfontein @briantist thanks for reviewing!

PR #223, opened 2023-03-10
collection_checklist: add clarification about version_added use

Andersson007:
##### SUMMARY
collection_checklist: add clarification about version_added use

I don't think any vote is needed as this info is present in https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#documentation-1

### DIFF
--- a/collection_checklist.md
+++ b/collection_checklist.md
@@ -28,7 +28,7 @@ Note for reviewers: If you don't know how to check any of the points below, plea
 - [ ] has `README.md`
 - [ ] documentation, examples, and return sections use FQCNs for the `M(..)` [format macros](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation) when referring to modules
 - [ ] modules (or plugins) from ansible-core use `ansible.builtin.` as a FQCN prefix
-- [ ] documentation and return sections use `version_added:` containing the *collection* version for which an option, module or plugin was added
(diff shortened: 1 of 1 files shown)


collection_checklist.md
```diff
@@ -28,7 +28,7 @@ Note for reviewers: If you don't know how to check any of the points below, plea
 - [ ] has `README.md`
 - [ ] documentation, examples, and return sections use FQCNs for the `M(..)` [format macros](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation) when referring to modules
 - [ ] modules (or plugins) from ansible-core use `ansible.builtin.` as a FQCN prefix
-- [ ] documentation and return sections use `version_added:` containing the *collection* version for which an option, module or plugin was added
+- [ ] documentation and return sections use `version_added:` containing the *collection* version for which an option, module or plugin was added (except cases when they were added in the very first release of the collection)
```
felixfontein, line 31: I think the original intent was "if `version_added` is there, it must use the collection version (and not the Ansible version, or some other random version)". (See also the RST file, it has a section "When using `version_added` in the documentation:" with the points in there that are summarized here.)

I now see that this point in the checklist can also be interpreted as "`version_added` must always be there".

I think your formulation improves this.

felixfontein (approved):

Andersson007:
@felixfontein thanks for reviewing, approving and clarifying!

